PR #63, opened 2018-07-04
Add object.format builtin function

### LINKED ISSUE #58: Add a built-in function "obj.format"
That takes an `IFormattable` and converts it to string using given format. Essentially I'm suggesting moving `math.format` to `obj.format` since it accepts all values and is not limited to numbers. Would also be nice if you could specify culture info.

Tyrrrz:
Closes #58

### DIFF
--- a/doc/builtins.md
+++ b/doc/builtins.md
@@ -1424,6 +1424,7 @@ The results of the multiplication: `value` * `with`
 Object functions available through the builtin object 'object'.
 
 - [`object.default`](#objectdefault)
+- [`object.format`](#objectformat)
 - [`object.has_key`](#objecthas_key)
 - [`object.has_value`](#objecthas_value)
 - [`object.keys`](#objectkeys)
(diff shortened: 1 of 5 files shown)


doc/builtins.md
```diff
@@ -1462,6 +1463,37 @@ The `default` value is returned if the input `value` is null or an empty string
 Yo
 ```
 
+[:top:](#builtins)
```
xoofx: Just to make sure, you did generate this by running the `Scriban.DocGen` program right?

Tyrrrz: Yep

Tyrrrz: It also generated docs for `date.now` but I removed it because it's not there in `master` branch. Also it makes the docs tests fail because it expects year 2017 in the output. I decided not to fix it since it's not related to this PR.

xoofx: ok, yep, will check for date.now

xoofx:
Thanks!

PR #76, opened 2018-10-02
Added Array.Contains functionality

marioantunes:
I'd like to contribute to this library as I'm in need of a contains method.


xoofx:
Thanks!

PR #77, opened 2018-10-05
Fix numeric literals not being parsed with the invariant culture

### LINKED ISSUE #74: Numbers in templates should be parsed using the InvariantCulture
Given the following template:

    {{ myVal = 1.0  }}
    myVal:  {{myVal}}
    {{ if myVal > 2 }}
       Parsed as ten
    {{ else }}
       Parsed as one
    {{ end }}`

The end result will depend on the culture. If I use a culture where the decimal separator is ",", this will print "parsed as ten".

reima:
Closes #74.


xoofx:
Thanks!

PR #79, opened 2018-10-05
Fix string to number conversion functions to use the context culture

### LINKED ISSUE #78: `string.to_(double|float|int|long)` do not use the template culture
The string to number conversion functions `string.to_double`, `string.to_float`, `string.to_int`, and `string.to_long` use the current thread culture to do the parsing. I think they should use the active `TemplateContext.CurrentCulture` instead.

reima:
Closes #78.


xoofx:
Thanks

PR #83, opened 2018-10-16
Add missing datetime formatting codes

### LINKED ISSUE #81: DateTime Format Milliseconds missing
I tried to format a datetime with milliseconds like this:
{{ date.format = '%d.%m.%Y - %H:%M:%S:%f'; testresult.time_stamp_send }}

It looks like that the 'f' placeholder is not supported :(

reima:
Closes #81.

I've implemented almost all the formatting codes mentioned in https://apidock.com/ruby/DateTime/strftime#864-Complete-Formatting-Codes and added test cases for them.

Formats that are not implemented are:
* `%3N`, `%6N`, `%9N`: Requires changes to the parsing logic to support multi-character codes. There are equivalent replacements available for two of those: `%3N` is `%L`, `%9N` is `%N`.
* `%z`: This should output the time zone offset from UTC (e.g. +0900). I left this out because `%Z` already outputs the time zone offset in a similar format (e.g. +09:00), even though it is documented to print the time zone name. I left it as it is for backwards compatibility.


reima:
Testing anything time zone related in Scriban is kind of tricky. `date.parse` returns a `DateTime` object with `DateTimeKind.Unspecified` or `DateTimeKind.Local`. So the output of the formating code `%Z` (time zone offset) depends on the time zone of the machine the test is running on. The issue with `%s` (UNIX timestamp in seconds) is similar: to get the seconds since 1970-01-01 midnight UTC the `DateTime` is converted to UTC. As the `DateTime` is interpreted as having the local time zone, this also depends on the machine's time zone.

I believe we should either
* remove the time zone dependent formatting codes (`%Z` and `%s`), as they can't be used in a meaningful way anyway, or
* look for another representation for date time values (DateTimeOffset from the BCL or OffsetDateTime/ZonedDateTime from Noda Time) which allow implementing them correctly

@xoofx What do you think?

xoofx:
> remove the time zone dependent formatting codes (%Z and %s), as they can't be used in a meaningful way anyway, or

I'm fine with this for now.

xoofx:
The tests are failing above:

````
1) Failed : Scriban.Tests.TestParser.400-builtins/440-date
  String lengths are both 286. Strings differ at index 213.
  Expected: "...PM\npm\n09:22:23 PM\n21:22\n1452025343\n23\n\t\n21:22:23\n2\n02\n 5-J..."
  But was:  "...PM\npm\n09:22:23 PM\n21:22\n1452028943\n23\n\t\n21:22:23\n2\n02\n 5-J..."
  ------------------------------------------------^
   at Scriban.Tests.TextAssert.AreEqual(String expectedValue, String actualValue, DiffStyle diffStyle, TextWriter output)
   at Scriban.Tests.TestParser.AssertTemplate(String expected, String input, Boolean isLiquid, Boolean isRoundtripTest, Boolean supportExactRoundtrip, Object model, Boolean specialLiquid)
   at Scriban.Tests.TestParser.TestFile(String inputName)
```
Could you have a look?

reima:
Yeah, that's the test that is failing because the AppVeyor test environment has a different time zone than my dev environment. Here's what happens:

https://github.com/lunet-io/scriban/blob/842026a8a48fe7ad120457fe19edc144f8eb4002/src/Scriban.Tests/TestFiles/400-builtins/440-date.txt#L7
Creates a `DateTime` instance with the specified values and a `Kind` of `DateTimeKind.Unspecified`.

https://github.com/lunet-io/scriban/blob/842026a8a48fe7ad120457fe19edc144f8eb4002/src/Scriban.Tests/TestFiles/400-builtins/440-date.txt#L36
Formats the `DateTime` as a UNIX timestamp. The implementation looks like this:

https://github.com/lunet-io/scriban/blob/842026a8a48fe7ad120457fe19edc144f8eb4002/src/Scriban/Functions/DateTimeFunctions.cs#L76

The UNIX timestamp is the number of seconds since 1970-01-01 00:00:00 **UTC**. To get that number the `dateTime` must be converted to UTC with `dateTime.ToUniversalTime()`. But the current time zone offset of `dateTime` is unspecified. In this case `ToUniversalTime` treats `dateTime` as a *local* time before converting to UTC. This is where the time zone setting of the test environment comes into play, and that's why the test fails on AppVeyor (but passes on my machine).

There are a couple of options to make this work:

* Implement a way to create `DateTime` instances with `DateTimeKind.Utc` and test with those (e.g. add a new Scriban function `date.parse_utc` that treats the date string as UTC).
* Remove the UNIX timestamp test from `440-date.txt`, acknowledging that the result is and should be time zone dependent. We could then add a separate test method that computes the expected time stamp according to the local time zone and compares it to the result of the formatting.

@xoofx Thoughts?

xoofx:
Let me know @reima if you need anything else from me to proceed with this PR (I haven't been able to follow this correctly recently)

Once this will be merged, I will probably release a new version of scriban

reima:
@xoofx Sorry for the delay, life has been busy. I've removed the failing test line, but haven't added a separate test method for the timestamp. Unfortunately I won't get around to that until next week. But I would also be fine with merging this PR as-is.

xoofx:
Thanks!

PR #87, opened 2018-10-29
add information about template name when path is null [because null p…

lafar6502:
add information about template name when path is null 
in my implementation of template resolver null path means there is no such template


xoofx:
Tests are failing, the test `TestTemplateLoaderIncludeWithNullGetPath` needs some upgrade 😉 

https://github.com/lunet-io/scriban/blob/bcbfdde8ac136181753c08c51b8abc79ea6e44a7/src/Scriban.Tests/TestIncludes.cs#L138-L146

xoofx:
Thanks!

PR #91, opened 2018-11-16
range for longs support

lafar6502:
(support long ranges + my previous commit because i'm uneducated in git and dont know how to make a proper pull request).


lafar6502:
cancel

PR #119, opened 2019-01-20
implements IgnoreHtmlNoise feature to the lexer

alex-becker-startp:
This feature is about enabling scriban templates to be edited with WYSIWIG HTML-Editors.
Using such an editor for editing leads to unparseable scriban blocks due to html code noise and unpredictable line breaks.

Consider the following simple sample:
```
This is a test for {{ 'scriban' }}: 
{{  a=5; 
    b = 3;
    a+b;
}}
```

The HTML-Code resulting from this input can easily be like the follwing:
```
<html>This is a test for {{ <span class="name">
'scriban'</span>&nbsp;}}: {{ &nbsp;a=5;<br> &nbsp;&nbsp;&nbsp;b 
= 3; 
 &nbsp;&nbsp;&nbsp;a
+b;
}}
```

obviously this will not be parseable by scriban at all.
But my changes will enable correct parsing when `LexerOptions.IgnoreHtmlNoise = true;` is set.

The following will apply within code-blocks which are not strings:
 - newlines within are ignored (parsed as whitespace)
 - HTML-Tags commonly used for newline are interpreted as newlines `<br>, </p>`
 - all other HTML-Tags are ignored (parsed as comment)
 - HTML-Entities (e.g. `&nbsp;`) are ignored (parsed as comment)

There are caveats which can be circumvented:
1) The html-opening tag (`<`) collides with the less-then operator.
 - when followed by a space `< ` or digit `<4` it will be interpreted as operator
2) When writing multi-line blocks, you should end every expression with a semi-colon.

One thing more:
I have absolutely minor experience with GIT. Please forgive me if I did something wrong posting this pull-request.

### DIFF
--- a/src/Scriban/Parsing/Lexer.cs
+++ b/src/Scriban/Parsing/Lexer.cs
@@ -461,6 +461,73 @@ namespace Scriban.Parsing
             }
         }
 
+        // Looks if following code is an html entity like &nbsp; or &#1234;
+        private bool PeekIsHtmlEntity()
+        {
+            int i = 1;
+            while (true)
+            {
+                var nc = PeekChar(i);
+
+                // check for entity end
+                if (nc == ';')
+                    return true;
+
+                // check for entity char
+                if (
+                    (nc == '#') ||
+                    (nc >= 'a' && nc <= 'z') ||
+                    (nc >= 'A' && nc <= 'Z') ||
+                    (nc >= '0' && nc <= '9')
+                    )
+                {
+                    i++;
+                    continue;
+                }
+
+                // invalid char;
+                return false;
+            }
+        }
+
+        // looks if the following code is a html tag like <tag> or <tag arg="value">
+        private bool PeekIsHtmlToken()
+        {
+            int i = 1;
+            bool isString = false;
+            while (true)
+            {
+                var nc = PeekChar(i);
+
+                // extra validate first char
+                if (i == 1)
+                {
+                    // exit when fist char is whitespace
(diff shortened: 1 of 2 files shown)


xoofx:
> This feature is about enabling scriban templates to be edited with WYSIWIG HTML-Editors.
Using such an editor for editing leads to unparseable scriban blocks due to html code noise and unpredictable line breaks.

Sorry, I'm going to decline this PR, you should probably have asked before, I don't want t to be rude here, just that...I don't want to bring such a change to the library. There are too many caveats, the interference with parsing can be significant, the implementation for correct tag parsing is more involving than what is done in these changes... and it feels overall a niche feature (that I would prefer not to have to support). 
This change is better off staying in your own repo/usage/application.

alex-becker-startp:
Okay, no problem.

What about a option that renders exception message to the output instead of stopping the render process?
This should not interference with anything and is super simple but useful.

```
This Block has an {{ 'error' | unknown_method }} !
```
resulting in
```
This Block has an [Pipe expression destination `unknown_method` is not a valid function] !
```

xoofx:
Why not. Maybe you could follow what was done for `TemplateContext.TryGetVariable`, instead provide a delegate `TemplateContext.TryCatchRuntimeException` with a delegate : `public delegate string TryCatchRuntimeExceptionDelegate(ScriptRuntimeException ex)`... no need to introduce a enable/disable, just check `TryCatchRuntimeException != null`

PR #130, opened 2019-02-08
Migrate all the project files to the new .NET Core csproj format et. al.

teo-tsirpanis:
For compatibility, they _still_ target .NET Framework 4.5.1.
Also, ignore files from Rider, move the solution file to the project's directory root (as is usually done), and add the word "Scriban" to Resharper's spell-check dictionary for this solution.


scriban.sln
```diff
@@ -3,32 +3,36 @@ Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
```
xoofx: Why the file was moved from src to top level? I would like to keep it where it was.

xoofx:
> Also, ignore files from Rider, move the solution file to the project's directory root (as is usually done),

For this, I would like that you revert this change. I have been adopting the sln in the src folder to keep the top folder smaller with only the required files.

scriban.sln
teo-tsirpanis: Done.

xoofx:
Thanks!

PR #132, opened 2019-02-08
Add base64 built-in functions.

### LINKED ISSUE #125: Built-in functions for base64 encoding/decoding

teo-tsirpanis:
My current implementation also adds functions to convert between strings and UTF-8 byte arrays. To be more explicit, a conversion to a UTF-8 byte array is needed before converting these bytes to base64, and vice versa.

This PR will close #125.


src/Scriban/Functions/StringFunctions.cs
```diff
@@ -971,10 +971,66 @@ public static string PadLeft(string text, int width)
...
         {
             return (text ?? string.Empty).PadRight(width);
         }
+
+        /// <summary>
+        ///  Encodes a string into an array of UTF-8-encoded bytes.
+        /// </summary>
+        /// <param name="text">The string to encode</param>
+        /// <returns>The encoded bytes</returns>
+        public static byte[] Utf8Encode(string text)
```
xoofx: this will require to regenerate the builtin delegates (run `Scriban.CodeGen` in the project)

teo-tsirpanis: It doesn't. Unless something is wrong... Wait a minute.

image

teo-tsirpanis:
Are you OK with the API? Another idea is to _assume_ the encoding to be UTF-8 and don't work with byte arrays at all.

I have no personal preference between either, actually.

src/Scriban/Functions/StringFunctions.cs
teo-tsirpanis: Is it normal? Should new code be generated? Did I break anything?

xoofx: I don't know,, maybe the problem is not handling this kind of parameter types, it is the first time that there is a `byte[]` so maybe there was an exception when you ran it?

teo-tsirpanis: No, there was no exception. The process returned 0.

xoofx:
> Are you OK with the API? Another idea is to _assume_ the encoding to be UTF-8 and don't work with byte arrays at all.

The problem is that byte arrays are completely new, and well, overall the API has been designed to manipulate strings... byte[] arrays are not really printable... so maybe you could remove it, because it is barely usable without exposing an object with an array... also, the type in itself (byte[]) might not be handled well across the runtime.

src/Scriban/Functions/StringFunctions.cs
xoofx: I don't know, you will have to debug, maybe something different that doesn't work well on your setup, or a bug in the program itself

teo-tsirpanis:
> it is barely usable

I was aware, but I also saw the array functions operating on `IEnumerable` and other collection interfaces. I will make conversions from strings to strings directly.

xoofx:
Thanks!

teo-tsirpanis:
The code generator has a bug I am investigating right now.

Can you merge this PR so that my next one builds on top of this one? Not all builtins are in `CustomFunction.Generated.cs` after all.

teo-tsirpanis:
> Can you merge this PR

Never mind, you already did. 😂

PR #134, opened 2019-02-09
Port unit tests to more frameworks et al.

### LINKED ISSUE #133: Does Scriban need to target all these frameworks?
Right now, Scriban targets:

* .NET Framework 3.5 (far too old, unsupported, does not run on anything .NET 4 cannot)
* .NET Framework 4 (unsupported, but OK if Windows XP support is needed)
* .NET Framework 4.5 (unsupported, redundant because of .NET 4 and .NET Standard 1.1, but OK because the regex timeout is a security feature)
* .NET Framework 4.7.1 (redundant because of .NET 4)
* .NET Standard 1.1
* .NET Standard 1.3 (OK, because it supports cryptographic hashing that .NET Standard 1.1 does not)
* .NET Standard 2.0 (redundant because of .NET Standard 1.3)
* .NET Core 2.0 (redundant because of .NET Standard 1.3)
* UWP (redundant because of .NET Standard 1.3, requires developers to download the entire UWP SDK)

My proposal is for Scriban to target only:

* .NET Framework 4
* .NET Framework 4.5
* .NET Standard 1.1
* .NET Standard 1.3

The proposal would simplify the project file, shrink the size of the NuGet package, make Scriban more maintainable.

teo-tsirpanis:
This PR:

* Closes #133 by targeting only the frameworks the maintainer specified.
* Ports the unit tests to .NET 3.5, .NET 4, and the older .NET Core versions that are still supported.
* Fixes the code generator to work with the new project format (the executable is one directory deeper).
* Adds [SPDX license metadata](https://github.com/NuGet/Home/wiki/Packaging-License-within-the-nupkg) to the project file.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -18,7 +18,7 @@ var template = Template.ParseLiquid("Hello {{name}}!");
 var result = template.Render(new { Name = "World" }); // => "Hello World!" 
 ```
 
-The language is very versatile, easy to read and use, similar to [liquid](http://liquidmarkup.org/) templates:
+The language is very versatile, easy to read and use, similar to [liquid](https://shopify.github.io/liquid/) templates:
 
 ```C#
 var template = Template.Parse(@"
@@ -89,13 +89,13 @@ Scriban is available as a NuGet package: [![NuGet](https://img.shields.io/nuget/
 
 Compatible with the following .NET framework profiles:
 
-- `.NET3.5`
-- `.NET4.0+`
--  .NET PCL profile `portable40-net40+sl5+win8+wp8+wpa81`
-- `UAP10.0+`
-- `NetStandard1.1+` and `NetStandard1.3+` running on `CoreCLR`
+- .NET Framework 3.5
+- .NET Framework 4.0
+- .NET Framework 4.5+ (supports asynchronous code and timeouts for regular expressions)
+- .NET Standard1.1+ (some features are not available)
+- .NET Standard1.3+ (which means .NET Core, Xamarin, UWP, Unity etc.)
 
-Also [Scriban.Signed](https://www.nuget.org/packages/Scriban.Signed/) NuGet package provides signed assemblies.
+Also the [Scriban.Signed](https://www.nuget.org/packages/Scriban.Signed/) NuGet package provides signed assemblies.
 
 ## Benchmarks
 
@@ -103,7 +103,7 @@ Also [Scriban.Signed](https://www.nuget.org/packages/Scriban.Signed/) NuGet pack
 
 ## License
 
-This software is released under the [BSD-Clause 2 license](http://opensource.org/licenses/BSD-2-Clause). 
+This software is released under the [BSD-Clause 2 license](https://opensource.org/licenses/BSD-2-Clause). 
 
 ## Related projects
 
--- a/src/Scriban.Tests/TestRuntime.cs
+++ b/src/Scriban.Tests/TestRuntime.cs
@@ -3,11 +3,16 @@
 
 using System;
 using System.Collections.Generic;
+#if !NETCOREAPP1_0 && !NETCOREAPP1_1
+using System.Data;
+#endif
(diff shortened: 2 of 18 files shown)


teo-tsirpanis:
Three tests are failing. Some JSON files differ by one space.

readme.md
```diff
@@ -89,21 +89,21 @@ Scriban is available as a NuGet package: [![NuGet](https://img.shields.io/nuget/
...
+Also the [Scriban.Signed](https://www.nuget.org/packages/Scriban.Signed/) NuGet package provides signed assemblies.
 
 ## Benchmarks
 
 **Scriban is blazing fast**! For more details, you can check the [benchmarks document](doc/benchmarks.md).
 
 ## License
 
-This software is released under the [BSD-Clause 2 license](http://opensource.org/licenses/BSD-2-Clause). 
+This software is released under the [BSD-Clause 2 license](https://opensource.org/licenses/BSD-2-Clause).
```
teo-tsirpanis: I changed most links to HTTPS. Your website's https is broken, so I left it as it was. I saw it is a GitHub Pages site, so it's not hard to issue a Let's Encrypt certificate.

src/Scriban/Scriban.csproj
```diff
@@ -9,80 +9,37 @@
...
+    <TargetFrameworks>net35;net40;net45;netstandard1.1;netstandard1.3;netstandard2.0</TargetFrameworks>
     <AssemblyName>Scriban</AssemblyName>
     <PackageId>Scriban</PackageId>
     <PackageId Condition="'$(SignAssembly)' == 'true'">Scriban.Signed</PackageId>
     <PackageTags>templating;handlebars;liquid</PackageTags>
     <PackageReleaseNotes>https://github.com/lunet-io/scriban/blob/master/changelog.md</PackageReleaseNotes>
     <PackageIconUrl>https://raw.githubusercontent.com/lunet-io/scriban/master/img/scriban.png</PackageIconUrl>
     <PackageProjectUrl>https://github.com/lunet-io/scriban</PackageProjectUrl>
-    <PackageLicenseUrl>https://github.com/lunet-io/scriban/blob/master/license.txt</PackageLicenseUrl>
-    <NetStandardImplicitPackageVersion Condition=" '$(TargetFramework)' == 'netstandard1.1' ">1.6.1</NetStandardImplicitPackageVersion>
```
teo-tsirpanis: What was that for? Should I have kept it?

src/Scriban.Tests/TestRuntime.cs
```diff
@@ -370,20 +378,21 @@ public void TestJson()
             var expected =
                 @"
 [
-  {
```
xoofx: Spaces have been modified while they should not

src/Scriban/Scriban.csproj
xoofx: This should be kept to make sure that the underlying version doesn't work

teo-tsirpanis: > to make sure that the underlying version doesn't work

I don't understand. I have no problem to remove it, but there must be a reason to explicitly the version of an infrastructure package. Were there problems in the past without this tag?

##### As a sidenote, AFAIK the version of this package does not have any relation with the version of .NET Standard the project targets.

xoofx: Can't remember exactly when, but I had this problem back in the days when they did upgrade the target version behind, so please, leave it as it is, it makes the build less machine/install dependent

src/Scriban.Tests/TestRuntime.cs
teo-tsirpanis: Done.

src/Scriban/Scriban.csproj
teo-tsirpanis: Done.

xoofx:
Thanks!

PR #146, opened 2019-03-11
Make StandardMemberRenamer.Rename accept a string.

teo-tsirpanis:
It makes testing it much easier by decoupling it from the hard-to-mock `MethodInfo`.

### DIFF
--- a/src/Scriban/Runtime/StandardMemberRenamer.cs
+++ b/src/Scriban/Runtime/StandardMemberRenamer.cs
@@ -1,5 +1,5 @@
 // Copyright (c) Alexandre Mutel. All rights reserved.
-// Licensed under the BSD-Clause 2 license. 
+// Licensed under the BSD-Clause 2 license.
 // See license.txt file in the project root for full license information.
 
 using System.Reflection;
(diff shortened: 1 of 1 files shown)


src/Scriban/Runtime/StandardMemberRenamer.cs
```diff
@@ -12,19 +11,18 @@ namespace Scriban.Runtime
...
 
         /// <summary>
-        /// Renames a camel/pascalcase member to a lowercase and `_` name. e.g `ThisIsAnExample` becomes `this_is_an_example`.
+        /// Renames a camel/pascalcase method name to a lowercase and `_` name. e.g `ThisIsAnExample` becomes `this_is_an_example`.
         /// </summary>
-        /// <param name="member">The member to rename</param>
+        /// <param name="name">The method name to rename</param>
         /// <returns>The member name renamed</returns>
-        public static string Rename(MemberInfo member)
+        public static string Rename(string name)
```
xoofx: Could you keep the previous implementation (call (string) from (memberInfo) ) so that it is not a breaking change

teo-tsirpanis: What if I create a new method called `DoRename` that accepts a string?

xoofx: As, it's just an overload of the function, it should keep the same name.

teo-tsirpanis: Even better.

xoofx: No it's not what I meant. StandardMemberRenamer.Rename(MemberInfo)  is a public API, so we need to keep both
StandardMemberRenamer.Rename(String) and the previous one, otherwise we would have to wait Scriban 3.0 to push this change

teo-tsirpanis:
That’s what I meant.

xoofx:
han sorry, as there was a resolved without an associated commit, I thought that you were reverting back.

xoofx:
(you should let the commit mark it as resolve instead of pushing the resolve button yourself, otherwise it is confusing)

teo-tsirpanis:
> (you should let the commit mark it as resolved instead of pushing the resolve button yourself, otherwise it is confusing)

How to do that BTW?

xoofx:
> How to do that BTW?

It should be automatic, but I'm not sure it is working with the force push. Better to leave the intermediate commits intact, I can choose to squash them just before merging

PR #147, opened 2019-03-12
Fix for appearance parser errors in ASP.NET Core developer exception page

tatarincev:
Continue to fix problem described in this PR https://github.com/lunet-io/scriban/pull/109


xoofx:
Thanks!

PR #169, opened 2019-05-18
Allow assignments only in top level expression statements

### LINKED ISSUE #164: Prevent assignments in boolean expressions
We've been using Scriban for a while in production (thanks!) and one (dangerous imo) pattern we've noticed is folks using a single "=" in "if" statements. Instead of `{{ if x == 2 }}` they write `{{ if x = 2 }}` which doesn't throw any exceptions and changes the values of `x` which causes confusion/bugs.

Is this something you'd be open to change? If yes, I could take a stab at a PR (any guidance as to where to start would be welcomed). Thanks.

pebezo:
This fixes #164 by making sure the Parse*Statement() functions that should allow only boolean expressions do not allow assignments.


pebezo:
@xoofx Any issue this with PR?

xoofx:
Oh no issue, just busy overall, on a business trip also this week, ping me back here next week and I will have a look!

xoofx:
It should be available in `2.1+`

PR #185, opened 2019-07-22
Case insensitive member lookup in ScriptObject

### LINKED ISSUE #184: Add support for case-insensitive ScriptObject
In some scenarios the templates will be authored by non-developers, and for them having to worry about the casing of member names is not natural.

Currently there's no easy way to do this ; the `ScriptObject.Store` property is initialized in the `ScriptObject` class and there's no way for external code to control it. It would be nice to have an overload of the `ScriptObject` constructor that accepts an `IEqualityComparer<string>`, in order to enable case-insensitive lookup of members.

thomaslevesque:
Fixes #184

### DIFF
--- a/src/Scriban.Tests/TestRuntime.cs
+++ b/src/Scriban.Tests/TestRuntime.cs
@@ -734,6 +734,18 @@ Tax: {{ 7 | match_tax }}";
             Assert.AreEqual("yo", result);
         }
 
+        [Test]
+        public void TestCaseInsensitiveLookupOnScriptObject()
+        {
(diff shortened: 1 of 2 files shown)


xoofx:
sorry, there is something wrong with coverall setup, but I'm gonna be busy for the next coming days, I will try to get back here in around 10 days... ping me back If I'm not ;)

thomaslevesque:
Sure @xoofx, no rush! In the meantime, I have a workaround anyway: I use `TryGetVariable` instead of pushing a `ScriptObject` on the `TemplateContext`.

PR #187, opened 2019-08-01
[WIP] Implement base ScriptVisitor

### LINKED ISSUE #186: Add a ScriptVisitor base class
As mentioned in https://github.com/lunet-io/scriban/issues/46#issuecomment-369517021, there's currently no visitor available to walk the AST. I recently needed one to extract variables and members from a template, so I rolled my own, but it would be nice if there was one built-in.

I can contribute my own implementation if you like.

thomaslevesque:
Fixes #186 

Marked as WIP because you'll probably have comments on the design.

### DIFF
--- a/src/Scriban/Syntax/ScriptNode.cs
+++ b/src/Scriban/Syntax/ScriptNode.cs
@@ -47,6 +47,8 @@ namespace Scriban.Syntax
         }
 
         public abstract void Write(TemplateRewriterContext context);
+
+        public abstract ScriptNode Accept(ScriptVisitor visitor);
     }
 
     public static class ScriptNodeExtensions
--- a/src/Scriban/Syntax/ScriptNode.cs
+++ b/src/Scriban/Syntax/ScriptNode.cs
@@ -48,7 +48,12 @@ namespace Scriban.Syntax
 
         public abstract void Write(TemplateRewriterContext context);
 
-        public abstract ScriptNode Accept(ScriptVisitor visitor);
+        public abstract void Accept(ScriptVisitor visitor);
+        public abstract TResult Accept<TResult>(ScriptVisitor<TResult> visitor);
+
+        public IEnumerable<ScriptNode> Children { get; }
+
+        protected abstract IEnumerable<ScriptNode> GetChildren();
     }
 
     public static class ScriptNodeExtensions
--- a/src/Scriban/Syntax/ScriptNode.cs
+++ b/src/Scriban/Syntax/ScriptNode.cs
@@ -104,6 +104,14 @@ namespace Scriban.Syntax
             }
             return false;
         }
+
+        public static TNode WithTriviaAndSpanFrom<TNode>(this TNode node, ScriptNode sourceNode)
+            where TNode : ScriptNode
+        {
+            node.Trivias = sourceNode?.Trivias;
+            node.Span = sourceNode?.Span ?? default;
+            return node;
+        }
     }
 
     public class ScriptTrivias
--- a/src/Scriban/Syntax/ScriptVisitor.cs
+++ b/src/Scriban/Syntax/ScriptVisitor.cs
@@ -14,111 +14,24 @@ namespace Scriban.Syntax
 
         public IScriptVisitorContext Context => _context;
 
-        public ScriptNode Visit(ScriptNode node)
+        private ScriptNode VisitCore(ScriptNode node)
         {
             if (node == null) return null;
 
             using (_context.Push(node))
             {
-                node = DispatchVisit(node);
+                node = node.Accept(this);
             }
 
             return node;
         }
 
-        protected virtual ScriptNode DispatchVisit(ScriptNode node)
+        public virtual ScriptNode Visit(ScriptTableRowStatement node)
         {
-            switch (node)
-            {
-                case ScriptTableRowStatement tableRowStatement:
-                    return VisitTableRowStatement(tableRowStatement);
-                case ScriptCaseStatement caseStatement:
-                    return VisitCaseStatement(caseStatement);
-                case ScriptElseStatement elseStatement:
-                    return VisitElseStatement(elseStatement);
-                case ScriptForStatement forStatement:
-                    return VisitForStatement(forStatement);
-                case ScriptIfStatement ifStatement:
-                    return VisitIfStatement(ifStatement);
-                case ScriptWhenStatement whenStatement:
-                    return VisitWhenStatement(whenStatement);
-                case ScriptWhileStatement whileStatement:
-                    return VisitWhileStatement(whileStatement);
-                case ScriptVariableGlobal variableGlobal:
-                    return VisitVariableGlobal(variableGlobal);
-                case ScriptVariableLocal variableLocal:
-                    return VisitVariableLocal(variableLocal);
-                case ScriptVariableLoop variableLoop:
-                    return VisitVariableLoop(variableLoop);
-                case ScriptArrayInitializerExpression arrayInitializerExpression:
-                    return VisitArrayInitializerExpression(arrayInitializerExpression);
-                case ScriptAssignExpression assignExpression:
-                    return VisitAssignExpression(assignExpression);
-                case ScriptBinaryExpression binaryExpression:
-                    return VisitBinaryExpression(binaryExpression);
-                case ScriptBlockStatement blockStatement:
-                    return VisitBlockStatement(blockStatement);
-                case ScriptCaptureStatement captureStatement:
-                    return VisitCaptureStatement(captureStatement);
-                case ScriptExpressionStatement expressionStatement:
-                    return VisitExpressionStatement(expressionStatement);
-                case ScriptFunctionCall functionCall:
-                    return VisitFunctionCall(functionCall);
-                case ScriptImportStatement importStatement:
-                    return VisitImportStatement(importStatement);
-                case ScriptIndexerExpression indexerExpression:
-                    return VisitIndexerExpression(indexerExpression);
-                case ScriptIsEmptyExpression isEmptyExpression:
-                    return VisitIsEmptyExpression(isEmptyExpression);
-                case ScriptMemberExpression memberExpression:
-                    return VisitMemberExpression(memberExpression);
-                case ScriptNamedArgument namedArgument:
-                    return VisitNamedArgument(namedArgument);
-                case ScriptNestedExpression nestedExpression:
-                    return VisitNestedExpression(nestedExpression);
-                case ScriptObjectInitializerExpression objectInitializerExpression:
-                    return VisitObjectInitializerExpression(objectInitializerExpression);
-                case ScriptPipeCall pipeCall:
-                    return VisitPipeCall(pipeCall);
-                case ScriptRawStatement rawStatement:
-                    return VisitRawStatement(rawStatement);
-                case ScriptReturnStatement returnStatement:
-                    return VisitReturnStatement(returnStatement);
-                case ScriptThisExpression thisExpression:
-                    return VisitThisExpression(thisExpression);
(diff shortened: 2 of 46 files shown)


src/Scriban/Syntax/ScriptNode.cs
```diff
@@ -104,6 +104,14 @@ public static bool HasTriviaEndOfStatement(this ScriptNode node, bool before)
...
             return false;
         }
+
+        public static TNode WithTriviaAndSpanFrom<TNode>(this TNode node, ScriptNode sourceNode)
+            where TNode : ScriptNode
+        {
+            node.Trivias = sourceNode?.Trivias;
+            node.Span = sourceNode?.Span ?? default;
+            return node;
+        }
```
thomaslevesque: Initially I had made this method internal, but when writing the tests I realized it could be useful to people who implement custom visitors.

src/Scriban/Syntax/ScriptVisitor.cs
```diff
@@ -0,0 +1,767 @@
...
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Scriban.Syntax
+{
+    public class ScriptVisitor
+    {
+        private readonly ScriptVisitorContext _context = new ScriptVisitorContext();
+
+        public IScriptVisitorContext Context => _context;
```
thomaslevesque: Visitor pattern implementations don't typically have that kind of thing, but I found it useful for my use case to be able to access the ancestors of the current node. It could go away if `ScriptNode` had a `Parent` property.

src/Scriban/Syntax/ScriptVisitor.cs
```diff
@@ -0,0 +1,767 @@
...
+
+            using (_context.Push(node))
+            {
+                node = DispatchVisit(node);
+            }
+
+            return node;
+        }
+
+        protected virtual ScriptNode DispatchVisit(ScriptNode node)
```
thomaslevesque: Virtual in case an implementer introduces new types of nodes (which I did in my use case)

src/Scriban/Syntax/ScriptVisitor.cs
```diff
@@ -0,0 +1,767 @@
...
+                return node;
+            }
+
+            var newStatement = new ScriptWhenStatement
+            {
+                Body = newBody,
+                Next = newNext
+            }.WithTriviaAndSpanFrom(node);
+
+            newStatement.Values.AddRange(newValues);
```
thomaslevesque: Any reason to make `Values` read-only? The code would be more efficient (and clearer) if it weren't. The nodes aren't immutable anyway, and collections are exposed as `List<T>`. Same is true of other node types.

(I assume the main benefit of making it read-only is that you can be sure it's never null)

src/Scriban/Syntax/ScriptVisitor.cs
```diff
@@ -0,0 +1,767 @@
...
+    {
+        private readonly ScriptVisitorContext _context = new ScriptVisitorContext();
+
+        public IScriptVisitorContext Context => _context;
+
+        public ScriptNode Visit(ScriptNode node)
+        {
+            if (node == null) return null;
+
+            using (_context.Push(node))
```
thomaslevesque: I know some people don't like this kind of "abuse" of `IDisposable`... Let me know if you prefer to change it to a try/finally instead.

xoofx:
Sorry to get back to you on this only today. That was during my summer holidays and then I completely forgot about it! 😅 
I'm not sure about the design... I would prefer a more conventional design with visitor pattern + Accept on the `ScriptNode`. ScriptNode would also provide a common abstract method for getting the number of child nodes and `GetChildrenAt` for example. Then a default visitor can use that to iterate on the nodes and call their Accept method.
`GetChildrenAt` would have to be implemented by each `ScriptNode` subclass (it is similar to Roslyn design).
I did implement this pattern for Tomlyn ([here](https://github.com/xoofx/Tomlyn/blob/7fd3bf419bf15fbbb80159004e4b4449bab99621/src/Tomlyn/Syntax/SyntaxNode.cs#L39-L59) and [SyntaxVisitor](https://github.com/xoofx/Tomlyn/blob/master/src/Tomlyn/Syntax/SyntaxVisitor.cs)

thomaslevesque:
Hi @xoofx,

Yes, I agree that a more conventional visitor implementation would be better. I did it this way because I wanted to use Scriban as-is, rather than use a modified version, so I couldn't change anything to the existing code.

I can work on this if you want, but I'm leaving on a long vacation soon, so it probably won't be before March.

What are your thoughts on my `IScriptVisitorContext`? I needed a way to access ancestors of the current node, but I'm not sure it's the best approach. An alternative would be to add a `Parent` property on `ScriptNode`, but it would have annoying consequences (e.g. an existing node couldn't be reused in a rewritten AST, because it would remove the link to its previous parent).

xoofx:
> I can work on this if you want, but I'm leaving on a long vacation soon, so it probably won't be before March.

No worries, no rush, do it at your convenience. I'm  not going to complain while I just looked at your PR 6 months later you submitted. 😅 

> What are your thoughts on my IScriptVisitorContext? I needed a way to access ancestors of the current node, but I'm not sure it's the best approach. An alternative would be to add a Parent property on ScriptNode, but it would have annoying consequences (e.g. an existing node couldn't be reused in a rewritten AST, because it would remove the link to its previous parent).

Yeah, for that, I don't know. This is something that have been asked already (getting the parent even outside of visiting). Ideally, this should have a proper parent property and then we would have to handle parenting on getter/setters (see Tomlyn for example), but that's a bit more laborious work...

thomaslevesque:
> Ideally, this should have a proper parent property and then we would have to handle parenting on getter/setters (see Tomlyn for example), but that's a bit more laborious work...

The trouble with this is that a visitor cannot reuse existing nodes to create new ones... Not ideal for performance.

thomaslevesque:
Hi @xoofx,

I changed the implementation to use double dispatch, as in the "canonical" visitor pattern. It still has two differences with the canonical pattern:
- The `Visit` methods return a `ScriptNode`, rather than `void`. This makes it easier to make a visitor that rewrites the AST.
- I kept the "context" bit, because it has been very useful to me. If you don't like it, I can drop it; if I make the `VisitCore` method virtual, it can be easily added by a derived implementation.

thomaslevesque:
> ScriptNode would also provide a common abstract method for getting the number of child nodes and GetChildrenAt for example. Then a default visitor can use that to iterate on the nodes and call their Accept method. Then a default visitor can use that to iterate on the nodes and call their Accept method.
> GetChildrenAt would have to be implemented by each ScriptNode subclass (it is similar to Roslyn design).

Somehow I missed that bit... I'll see if I can add this. It sure would make the whole thing simpler.

thomaslevesque:
OK, in fact it's going to be difficult to implement a `DefaultVisit`: because of the ability to rewrite the AST, the visitor needs to be able to recreate a node with different children. So I can't just work with abstract children, I need to know *which* children they are.

Of course, I could drop the whole AST rewriting capability, but it would make the feature less useful to me.

xoofx:
> Somehow I missed that bit... I'll see if I can add this. It sure would make the whole thing simpler.

Yeah, it is usually easier to work like this, as it put the implementation detail in the class. But this visitor doesn't return new node.

> OK, in fact it's going to be difficult to implement a DefaultVisit: because of the ability to rewrite the AST, the visitor needs to be able to recreate a node with different children. So I can't just work with abstract children, I need to know which children they are

So you can't reuse the same implementation than the visit above. For example, Roslyn is generating these rewriter 
[https://github.com/dotnet/roslyn/blob/8017891b5b8174526ba43a8bb46638c2a8618626/src/Compilers/CSharp/Portable/Generated/Syntax.xml.Main.Generated.cs#L1310-L1956] which inherits `CSharpSyntaxVisitor<TResult>`.  

While for `CSharpSyntaxVisitor.DefaultVisit` it is following [here](https://github.com/dotnet/roslyn/blob/8017891b5b8174526ba43a8bb46638c2a8618626/src/Compilers/CSharp/Portable/Syntax/CSharpSyntaxWalker.cs#L37) the general method.

So you would have to write similar code to what you wrote in the Visitor. Just that it is a particular visitor replacing node, node just visiting or even creating new type of results (e.g not nodes).

It's ok to have a `ScriptVisitor`, a `ScriptVisitor<TResult>` and a `ScriptRewriter` (which would be your actual `ScriptVisitor`)

thomaslevesque:
I'm not sure I understand... What would you like me to do exactly? Write separate visitor and rewriter classes?

xoofx:
> I'm not sure I understand... What would you like me to do exactly? Write separate visitor and rewriter classes?

Yeah, they can't be written in the same class, it's very similar to Roslyn. You will have 3 classes:
- `ScriptVisitor` which can use a `DefaultVisi()` implementation that would rely on  `ScriptNode.GetChildrenCount` / `ScriptNode.GetChildrenAt` from `ScriptNode`
- `ScriptVisitor<TResult>`  which would have a `DefaultVisit` empty
- `ScriptRewriter` inheriting from `ScriptVisitor<ScriptNode>` which is actually similar to what you have written, implementing the Visit methods.

It's not very far from what you have. Though, the additional work is to provide `ScriptNode.GetChildrenCount`/`ScriptNode.GetChildrenAt` and their derived implementations for each `ScriptNode` implem.

thomaslevesque:
Ah, I see.
Not sure I see the point of `ScriptVisitor<TResult>`, though... The default implementation is useless, and if you ever want to make an implementation than returns anything other than `ScriptNode`, you need to redo all the work of walking the child nodes. Do you have a specific use case in mind?

thomaslevesque:
BTW, is there a particular reason to expose `ChildrenCount`/`GetChildrenAt` rather than just a `IEnumerable<ScriptNode> Children` property? It would be (slightly) less painful to implement. Is this for performance reasons?

xoofx:
> Not sure I see the point of ScriptVisitor<TResult>, though... The default implementation is useless, and if you ever want to make an implementation than returns anything other than ScriptNode, you need to redo all the work of walking the child nodes. Do you have a specific use case in mind?

No personal use case right now for `ScriptVisitor<TResult>`, it's ok not to introduce it as it would not break the API.

> BTW, is there a particular reason to expose ChildrenCount/GetChildrenAt rather than just a IEnumerable<ScriptNode> Children property? It would be (slightly) less painful to implement. Is this for performance reasons?

Yeah, but that's ok to expose only a `Children` property for now and use `yield`. We can add `ChildrenCount`/`GetChildrenAt` later if necessary

thomaslevesque:
I eventually implemented the `ScriptVisitor<TResult>`, because it impacts the signature of the `ScriptNode.Accept` method, so it would have been a bit messy to add it later. And this way there's not need for an `Accept(ScriptRewriter)` method.

I renamed the existing tests to `ScriptRewriterTests`, but I didn't add tests for the new `ScriptVisitor` yet. I can't think of a good approach to test it. Any idea?

xoofx:
I'm gonna probably pull your PR into the new branch [devel-3.0](https://github.com/lunet-io/scriban/tree/devel-3.0) as I going to need a rewriter in that branch, and as it is a new version allowing breaking change API, I will see if there is anything that I could improve with your work. Thanks a lot!

xoofx:
Closing this branch as your PR is merged into devel-3.0. If you want to iterate on it, feel free to open a PR on this branch.

thomaslevesque:
> I'm gonna probably pull your PR into the new branch [devel-3.0](https://github.com/lunet-io/scriban/tree/devel-3.0) as I going to need a rewriter in that branch, and as it is a new version allowing breaking change API, I will see if there is anything that I could improve with your work. Thanks a lot!

Fine by me, thanks!

PR #220, opened 2020-01-27
Fix markdown syntax in runtime.md

yekalkan:


xoofx:
Thanks, replicated the fix into the commit cacfc1cce514f407466e6460c1522e14fe583ddd

PR #226, opened 2020-04-06
Bundle the package icon with the nupkg file.

teo-tsirpanis:
And some other minor tweaks.

PR #229, opened 2020-04-20
Reference net35 libraries only when required and make development libraries' assets private

### LINKED ISSUE #228: NuGet dependencies
Hi,

I recently upgrade to v2.1.2 (from v1.x) and notice that a number of NuGet dependencies are also included. I'm not convinced that they're all necessary. For example, my project is a `net472` project and hence references your `net45` build. which has a dependency for `jnm2.ReferenceAssemblies.net35` and `Microsoft.NETFramework.ReferenceAssemblies`.

From what I gather, both of these dependencies are required only for *building* Scriban and it's not necessary to deploy the dependencies, i.e. set `privateAssets="true"` in the `<PackageReference>` element. 

Further, `jnm2.ReferenceAssemblies.net35` is required only for building the `net35` target, right?

And on a slightly related note, are separate `netstandard1.1` and `netstandard1.3` libraries necessary? I see they both reference the `NETStandard.Library` v1.6.1, so it looks a bit weird. I'm not 100% sure about this though.

igitur:
Fixes #228 

I see this is also partially addressed in #226 , but there are some other diffs in that PR that seem non-related.

### DIFF
--- a/src/Scriban/Scriban.props
+++ b/src/Scriban/Scriban.props
@@ -29,9 +29,11 @@
   <ItemGroup Condition=" '$(TargetFramework)' == 'net35'">
     <Reference Include="System.Web" />
   </ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net35'">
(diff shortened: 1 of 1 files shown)


xoofx:
Thank you!

PR #230, opened 2020-04-23
Scriban 3.0

xoofx:
This is a draft PR for next version of Scriban, still wip:

- [x] Remove support for NET FWK 3.5, 4.0. Support only for `NET Standard 2.0+`
- [x] Add support for a new scientific parsing mode (different from default Scriban).
- [x] Add support for hexa/binary and big integers
- [x] Add support for custom unary/binary operators on types
- [x] Add support for `ScriptVisitor` and `ScriptRewriter` API to visit and rewrites syntax trees.
- [x] Improve roundtrip parsing by changing how code/escape enter/exit and end statements were parsed (now fully parsed not as trivias).
- [x] API is breaking in many places, but the parsing/compatibility of existing scripts should not change (apart for the case of `<<`, TBD if we really want to make is breaking or not or keep it only for scientific mode)
- [x] Cleanup `IScriptCustomFunction` interface (changed, but not yet final)
- [x] Add `ScriptFormatter`  API to allow to reformat a script (mostly useful for script only mode)
- [x] Add tests for scientific mode
- [x] Improve performance after previous additions
- [x] Add support for more binary expressions for ScriptArray
- [x] Add support for conditional ternary expression (e.g `condition ? thenValue : elseValue`)
- [ ] Add decimal support by default with overflow to double
- [ ] Update documentation about scripting mode and API changes. Update main readme. Remove strict notion that Scriban is only a text templating engine but can be used also as a simple scripting engine.


xoofx:
@thomaslevesque so this PR is integrating your original PR for script rewriting. As part of Scriban 3.0, I wanted to support a full syntax tree with correct parenting (which means that now you can't have a node reused in another or create cycles) but I found laborious to update manually all these methods ( ChildrenCount, GetChildren, Accept...etc.), so the rewriter and all these methods are now generated as part of a `Scriban.AsyncCodeGen` (will probably rename it). But overall the experience is amazing, all this boilerplate code is generated automatically now.

thomaslevesque:
> As part of Scriban 3.0, I wanted to support a full syntax tree with correct parenting (which means that now you can't have a node reused in another or create cycles)

I kinda liked being able to reuse a node without changing it, but it's true that having the parent accessible directly on the node is useful, and it removes the need for my "context" thing.

> the rewriter and all these methods are now generated as part of a `Scriban.AsyncCodeGen` (will probably rename it). But overall the experience is amazing, all this boilerplate code is generated automatically now.

Nice!

xoofx:
> I kinda liked being able to reuse a node without changing it, but it's true that having the parent accessible directly on the node is useful, and it removes the need for my "context" thing.

Yeah, I agree, it was more easy and go 🙂 , I was worried taking that route that I would have lots of stuffs to fix but it went ok and the cases where you need to detach a node was fairly uncommon in the code base. But it helps to prevent mistakes and yeah, you get the parenting for free, which will mean for example that in a `IScriptCustomFunction` it will be able to provide functions in a contextual context, which is a niche feature but could become very handy in some cases.

PR #257, opened 2020-10-16
fix rendering of conditional expression in table

Guria:
Expression `||` were treated as table column delimiter so it should be escaped to be rendered as expected.
Fixed spelling, added note that conditional expressions always returns boolean but not its members.


Guria:
@xoofx  not sure why ci is failing

xoofx:
Thanks!

PR #265, opened 2020-11-28
Documentation: Fixes typo on array builtin functions examples.

rroa:
* Updates documentation for array builtins. Fixes typo on add_range example.

Behavior has been tested to ensure it matches this updated example.


xoofx:
Thanks, this is fixed by commit 54e720a (the code has to be changed instead, and run DocGen, but I won't blame you as it is not explained anywhere in the docs!)

PR #272, opened 2020-12-11
Scriban SourceGenerator

MarkPflug:
This PR is relating to an exchange I had on twitter with @xoofx earlier today. This is a functioning WIP source generator package for using Scriban as a code gen template language akin to T4.

Add the package to your project, create a Scriban text template .stt (open to a better suggestion) file in the project and the output will be included in the compilation. Obviously, the generated code won't be included in the project source. Generated code shows up in Intellisense. Should work in any build environment, thought VS Code might not have functioning intellisense.

Better than T4:
- Generated code not in source.
- Doesn't require VS or VSCode extension.
- Scriban is (presumably) superior to T4.
- Doesn't pollute the .csproj with T4 wiring.

This would still need some work to provide good diagnostic output when the template has errors. I'm actually not familiar with the Scriban language though, so I need to familiarize myself to do that portion of the work.

Might be useful to have a second file format for providing template parameters. Right now each template is processed without parameters, expecting it to be self-contained.

Anyway, the question is: would you be interested in merging this into your project? Is it worth me investing the time in providing diagnostics?


xoofx:
Hehe, thanks for trying that 😃 

> Anyway, the question is: would you be interested in merging this into your project? Is it worth me investing the time in providing diagnostics?

Good question! So at first, I was thinking that it would not need to be part of this repo, it could make sense so that it is in sync with any fixes that would come up with Scriban, but it's a problem in itself: If you want to release an update on the side for the generator, it would be more complicated... so...hm... maybe a separate project on your side would be better. But I can definitely give a PR review before you take it to another repo?

> Scriban is (presumably) superior to T4.

I would not say superior. It's different. 😉  For example, Scriban doesn't give access to any .NET API, so you can't do anything you want (e.g access network). Functions exposed by Scriban are in a controlled execution environment. So it can be feel more limited. For the case of source generators, it's not a bad thing, because you want your source generators to not depend on anything that Roslyn is not aware of.

Otherwise, as I mentioned on Twitter, it can be bad to have such code generator in your Roslyn compiler, specially if your templates generates thousands of lines of code that are systematically the same for each compilation. It will make Roslyn compiler run more slowly. So that's probably something that will have to be stated quite clearly.

> This would still need some work to provide good diagnostic output when the template has errors. I'm actually not familiar with the Scriban language though, so I need to familiarize myself to do that portion of the work.

First look at your PR, you report the Scriban errors and it should be probably enough.

MarkPflug:
It made sense to me to live in this repository because it would likely want to be released at the same cadence as the Scriban base library. I guess that doesn't dictate that it lives here, but it seemed like it would be easier.

You might be right about the SourceGenerators being a bad fit for this. My interest in working on this is to have a real-world example of using a SourceGenerator. But, since this doesn't need to interrogate the syntax of the source compilation, it isn't a particularly interesting example of what SourceGenerators are capable of. This *could* be entirely implemented as an MSBuild nuget package.

> It will make Roslyn compiler run more slowly

I was under the impression that the SourceGenerator engine had some mechanism for detecting when a generated source was out of date, and reusing cached syntax trees otherwise. However, looking at the shape of the interfaces I might be wrong about that. I think, perhaps, the source generator will be run every time, but if it produces identical source text then Roslyn will reuse a cached SyntaxTree from that compilation. It might be possible to achieve similar caching of the generated source when the input AdditionalText (the .stt file) hasn't changed. Then, only when one of the template files is modified will any expensive work need to be done that will flow through the entire compilation phase.

> Scriban doesn't give access to any .NET API, so you can't do anything you want (e.g access network).

I would argue that makes it superior to T4, at least for how sane people use T4.

My other motivation for wanting this to live in this repository, and have the Scriban name attached, is that it would make it more discoverable for the community. No one is going to find or use a package that I produce on my own, I don't have the marketing clout.

Anyway, I plan on continuing to tinker with this. If you'd rather I move it into a separate project just let me know.

Jogai:
Time for a scriban github org?

MarkPflug:
@xoofx About your performance concern: I've investigating caching at the SourceGenerator level, and it does currently appear to be lacking a bit. Performance of the IDE is somewhat impacted when generating large source files (I tested with ~2MB output). I asked the roslyn guys about this, and it sounds like this is intended to be a supported scenario are working on some improvements to allow better caching and reuse:
https://github.com/dotnet/roslyn/discussions/49921#discussioncomment-195390

Anyway, I'm pretty excited with what I'm seeing. Using Scriban as C# templating language feels pretty pleasant to me, at least compared to my memories of working with T4.

xoofx:
Thinking about it, one potential issue I can see for now is that afaik, you can't reference other assemblies easily with source generators, so you would have to [ILMerge](https://github.com/dotnet/ILMerge) it.

badcel:
>you can't reference other assemblies easily with source generators

@xoofx: do you mean that you would need to do this? [Roslyn Cookbook](https://github.com/dotnet/roslyn/blob/master/docs/features/source-generators.cookbook.md#use-functionality-from-nuget-packages)

MarkPflug:
Not sure I follow? You mean I'd have to ILMerge the scriban library into my source generator?
It currently works with scriban as a separate assembly carried in the same folder as the source generator assembly; this works *in my environment*. You might be right that it won't work everywhere. I've run into a issues with other tools where things worked running as net core, but then have issues when running as net framework. Not sure if this is still true as both the compiler, MSBuild, and the dotnet cli have evolved since I had those issues.

xoofx:
> @xoofx: do you mean that you would need to do this? [Roslyn Cookbook](https://github.com/dotnet/roslyn/blob/master/docs/features/source-generators.cookbook.md#use-functionality-from-nuget-packages)

Yep exactly, nice that they have already a cookbook for this case

MarkPflug:
I think it  might be okay though, because scriban targets netstandard2.0 which should work in both core and framework.

xoofx:
> I think it might be okay though, because scriban targets netstandard2.0 which should work in both core and framework.

Yes.

So if an assembly on the side is working (as per the cookbook), I'm not sure they use Assembly Load Context to avoid clashes with between analyzers loaded at runtime... hence an ILMerge might be the "safest" solution

MarkPflug:
"runtime", you mean the runtime of the compiler? Wouldn't that only be a problem if some other analyzer/sourcegenerator also tried to load an incompatible version of scriban? Not impossible, but unlikely.

xoofx:
> "runtime", you mean the runtime of the compiler? Wouldn't that only be a problem if some other analyzer/sourcegenerator also tried to load an incompatible version of scriban? Not impossible, but unlikely.

Well, the case of Newtonsoft.Json is a typical one. I have seen other folks trying to use scriban in source generators as well, so it can be quite common...

xoofx:
Example: https://www.cazzulino.com/source-generators.html

xoofx:
See also this discussion https://github.com/dotnet/roslyn/issues/49075#issuecomment-722027846  with @PathogenDavid

MarkPflug:
I see. People are using scriban to emit the code rather than using a StringBuilder approach which is very tedious and hard to maintain. That makes sense, and you're right, this will probably be more common than I thought.

xoofx:
> I see. People are using scriban to emit the code rather than using a StringBuilder approach which is very tedious and hard to maintain. That makes sense, and you're right, this will probably be more common than I thought.

Yeah, I'm sure it will become more a problem for source generators than with analyzers where most analyzers don't need a dependency.

MarkPflug:
This might be another reason why it would be easier to have the source generator in the same repository: I could just statically include all the scriban source <Compile Include="../scriban/**/*.cs"/>. No ILMerge needed.

xoofx:
> This might be another reason why it would be easier to have the source generator in the same repository: I could just statically include all the scriban source . No ILMerge needed.

Indeed 😉

xoofx:
Though you can still do that with a git submodule 😅

MarkPflug:
I guess. But then I'd have to learn what a git submodule is. 😆

MarkPflug:
Okay, the static include approach works. Amazingly trivial, just needed to exclude a couple files: assemblyinfo.cs.

PathogenDavid:
> I guess. But then I'd have to learn what a git submodule is. 😆

In my experience: They're a quirky Git feature that once you use them too much you get addicted and annoy everyone who hates them conceptually or isn't familiar with them and you end up having to remind people to run `git submodule update --init --recursive` on a regular basis. 8/10 would recommend.

In seriousness: They're a useful tool, and I'd recommend learning how to use them. Essentially they let you reference external Git repos from yours. [Here's an example of them in use](https://github.com/InfectedLibraries/InfectedImGui/tree/main/external).

Git subtree is another similar concept with a different set of tradeoffs. Can't say I've really used them or even seen them used in practice though.

> This might be another reason why it would be easier to have the source generator in the same repository: I could just statically include all the scriban source . No ILMerge needed.

Another option is distributing a NuGet source package for Scriban. They aren't common, but they do exist. [xunit.assert.source](https://www.nuget.org/packages/xunit.assert.source/) is an example that comes to mind. ([nuspec here](https://github.com/xunit/xunit/blob/a8f55d61022f39362ff10c1f73eab0cb965c2619/src/xunit.v3.assert/xunit.v3.assert.source.nuspec))

badcel:
> Another option is distributing a NuGet source package for Scriban.

I would like to see this, as this makes life easier for all people who want to use Scriban in a source generator :+1:  (including me).

Currently I create a nuget package from my source generator on build and include it via a `PackageReference` in my second project (via a local nuget folder) although both projects are in the same repository. I see no other way to include Scriban cleanly in my source generator. Except perhaps for ILMerge, which I just learned that it exists.

I think source nugets could improve the use case of Scriban in source generators quite a bit.

xoofx:
> Another option is distributing a NuGet source package for Scriban. They aren't common, but they do exist. xunit.assert.source is an example that comes to mind. (nuspec here)

Indeed, I could do something similar, likely even as part of the same package without having to ship a separate binary/source package and with types all internal. Let me check...

MarkPflug:
I think @PathogenDavid is right: I think including scriban as a git submodule would be better than as a source package.

For one, you'd have to do some unfortunate code-splitting (partial class) or `#if` hackery to publish a library (public) and source package (internal) out of the same repo. I mean, it's your project, so do what you will, but I wouldn't want to deal with that. Or, maybe there is some other trick that I'm not aware of.

xoofx:
> For one, you'd have to do some unfortunate code-splitting (partial class) or #if hackery to publish a library (public) and source package (internal) out of the same repo. I mean, it's your project, so do what you will, but I wouldn't want to deal with that. Or, maybe there is some other trick that I'm not aware of.

I have a solution working right now with a single package that will work for everyone. Should be part of 3.2.0

xoofx:
Hm, things are actually more complicated, because Scriban has other dependencies (e.g `ValueTask<T>` ...etc.)... I will have to figure out how I can workaround that...

xoofx:
As part of the PR #281 I have added support for embedding sources directly with the Scriban package, it's part of 3.2.0+

xoofx:
> As part of the PR #281 I have added support for embedding sources directly with the Scriban package, it's part of 3.2.0+

Actually, it will be in 3.2.1, I forgot to add the props/targets files to git for 3.2.0 😅

caesay:
Hey, I know this is an old issue but I'm a bit confused as to the result of a "Scriban source generator". A modern alternative to T4 (in the context of generated code to be included in the build) would be amazing and a roslyn source generator capable of transforming scriban templates would solve this. The linked PR #281 seems to add support for bundling the Scriban source code into your project, but not transforming scriban templates on build? 

So was this source generator actually built and/or relocated somewhere else?

Issue #279, opened 2020-12-20
Abstract Syntax Tree : span of a parent node is smaller than the sum of spans of its children


NeVeSpl:
For the given template:
```{{filePath = "result\\" | string.Append class.Name }}```
we get following syntax tree:
![image](https://user-images.githubusercontent.com/11632757/102719548-0a28e000-42ef-11eb-85eb-c24453b8b3df.png)

where in brackets are printed out offsets of span :  ({node.Span.Start.Offset} - {node.Span.End.Offset}):

It is a bug, or by design?

xoofx:
It is likely a bug, but I don't have personal time to investigate this issue, feel free to investigate it.

PR #280, opened 2020-12-20
AST : fix parent node span does not include all its children

NeVeSpl:
Solution for problems reported in #279.

Also as a bonus, it would be nice to have Length property on SourceSpan struct, which was added in last check-in.


xoofx:
Thanks!

> Also as a bonus, it would be nice to have Length property on SourceSpan struct, which was added in last check-in.

That will require to bump to 3.1. Might have to check other pending issues/bugs before releasing a new versions.

PR #286, opened 2020-12-23
Disable nullability warnings

### LINKED ISSUE #285: Using nuget source package with enabled nullables creates warnings
In general the new nuget source package works fine. Thank you for your work. Now it is very easy to depend on Scriban in a source generator.

One small hickup is if you use the package in a project which enables the non nullable reference types feature there are ~600 warnings regarding nullables.

Now it is quite hard for me to differentiate between my own nullable warnings and the ones from Scriban.

badcel:
Fixes #285.

### DIFF
--- a/src/Scriban/Functions/ArrayFunctions.cs
+++ b/src/Scriban/Functions/ArrayFunctions.cs
@@ -2,6 +2,8 @@
 // Licensed under the BSD-Clause 2 license.
 // See license.txt file in the project root for full license information.
 
+#nullable disable
+
 using System;
 using System.Collections;
 using System.Collections.Generic;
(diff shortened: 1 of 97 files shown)


badcel:
One remark: I only added the code in files which actually produce warnings. Perhaps the nullability feature should be turned on for the whole project and warnings should be treated as errors (in general or only for the nullabillity feature) to be aware of it in case some code changes.

xoofx:
Thanks!

badcel:
@xoofx what do you think of my remark? This could be done with an entry in the `.editorconfig` like

```
[*.cs]
dotnet_diagnostic.CS8618.severity = error
```

xoofx:
I have pushed a commit 3c60ad0c969ac89b27282483e15c024ad020b428 that adds TreatWarningsAsErrors

badcel:
Great 👍 

To actually get the nullabillity errors the non nullable reference types needs to be activated in csproj, too. Sorry for still pinging you here 🙈.

xoofx:
Oh, yeah, I forgot, sorry

PR #303, opened 2021-01-14
fix for Scriban.DelegateCodeGen

hzdgopher:
1.
changed "_methods" in Scriban.DelegateCodeGen/Program.cs
from: 
    Dictionary<string, MethodDefinition>
to:
    Dictionary<string, List<MethodDefinition>>
to allow the storage of multiple methods with the same signature.

2. 
changed MethodComparer in Scriban/Runtime/DynamicCustomFunctions.cs
to make equality fit the above "same signature but different method/type" logic

3.
the resulting new default in Scriban/Runtime/CustomFunction.Generated.cs


xoofx:
Thanks!

Issue #128, opened 2019-02-05
Add `string.IsNullOrWhiteSpace`


dorfire:

xoofx:
Indeed, PR welcome

n1l:
@xoofx https://github.com/scriban/scriban/pull/307

AlfredoMS:
@xoofx looks like this issue can be closed anytime now

PR #307, opened 2021-01-18
PR for Add `string.IsNullOrWhiteSpace` #128

n1l:
I think I understood the issue correctly. Please review.

### DIFF
--- a/doc/builtins.md
+++ b/doc/builtins.md
@@ -2311,6 +2311,66 @@ true
 ```
 
 [:top:](#builtins)
+### `string.empty`
+
+```
+string.empty <text>
+```
+
+#### Description
+
+ Returns a boolean indicating whether the input string is an empty string.
+
+#### Arguments
+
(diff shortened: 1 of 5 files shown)


src/Scriban/Syntax/Expressions/ScriptBinaryOperator.cs
```diff
@@ -191,6 +194,10 @@ public static string ToText(this ScriptBinaryOperator op)
                     return "&";
                 case ScriptBinaryOperator.BinaryOr:
                     return "|";
+                case ScriptBinaryOperator.IsEmpty:
```
n1l, line 197: I think that not correct. But I'm not sure where should I include the function. LiquidBuiltinsFunctions class, TryLiquidToScriban method?
I doubt that there are anything like whitespace of empty in liquid language. So that the question.

xoofx:
Nice PR, thanks!

PR #309, opened 2021-01-19
always return AST, even if it contains errors

NeVeSpl:
It would be nice to have access to AST even if is not entirely correct. Anything is better than null/nothing.


xoofx:
Thanks

PR #321, opened 2021-01-26
Basic investigation of exception thrown during evaluate process.

n1l:
It's not a completed code, it's more like a question.

I found a reason I think, for the latest loop step, Current node is set to null but the node is mandatory context for all internal array functions.
I've added a code to demonstrate that.

Do you have anything to suggest what else I can check?

### DIFF
--- a/src/Scriban.Tests/TestParser.cs
+++ b/src/Scriban.Tests/TestParser.cs
@@ -78,6 +78,21 @@ namespace Scriban.Tests
         }
 
         [Test]
+        public void TestEvaluateProcessing()
+        {
+            {
+                var result = Template.Parse("{{['', '200', '','400'] | array.filter @string.strip}}").Evaluate(new TemplateContext());
+
+                Assert.AreEqual(new[] { "", "200", "", "400" }, result);
+            }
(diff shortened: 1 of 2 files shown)


xoofx:
> I found a reason I think, for the latest loop step, Current node is set to null but the node is mandatory context for all internal array functions.

When/Where does it happen? I believe a proper fix would to avoid a dangling null node in the middle of eval, so if there is anything setting it to null, we should probably avoid setting null.

n1l:
As I said it's not a completed code.

The exception throws when we are trying to get a result. Here:
image

I'm not sure what the correct behavior should be here. So I'm asking what else I can check.

xoofx:
> I'm not sure what the correct behavior should be here. So I'm asking what else I can check.

That's what I suggested. Could you check when the node is going back to null? (e.g breakpoint/debug on the setter)

NeilMacMullen:
I think the issue is the same as explained in #322 - at least, the fix for that also allows the unit test in this PR to pass.

NeilMacMullen:
Fixed by changes in #324

PR #326, opened 2021-01-31
Implement some missing interface methods.

teo-tsirpanis:
ScriptObject got a more efficient implementation of ``IEnumerator`2.GetEnumerator``, a correct implementation of ``ICollection`1.IsReadOnly`` and non-throwing implementations of four ``ICollection`1`` members (`CopyTo` in particular is used by Rider's debugger and hindered debugging experience by throwing).
Furthermore `ScriptRange`'s `IList.Remove` was trivially implemented.


xoofx:
Thanks!

Issue #331, opened 2021-02-15
Feature request: Built in random and Guid functions


CommonGuy:
It would be nice to have built in functions to generate random numbers and GUIDs/UUIDs.

Would a PR that implements this be accepted? Would probably have to define the "prefix" (math? object? other?) for the built in functions first.

xoofx:
Yep, they can go in `math` probably

PR #332, opened 2021-02-16
Add math.random and math.uuid built in functions

CommonGuy:
Implemented math.random and math.uuid_v4 according to https://github.com/scriban/scriban/issues/331

### DIFF
--- a/src/Scriban/Functions/MathFunctions.cs
+++ b/src/Scriban/Functions/MathFunctions.cs
@@ -25,6 +25,8 @@ namespace Scriban.Functions
 #endif
     partial class MathFunctions : ScriptObject
     {
+        private const string RandomTagKey = "random";
+
         /// <summary>
         /// Returns the absolute value of a specified number.
         /// </summary>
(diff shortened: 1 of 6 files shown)


src/Scriban/Functions/MathFunctions.cs
```diff
@@ -335,5 +337,48 @@ public static object Times(TemplateContext context, SourceSpan span, object valu
...
+        /// <returns>The created UUID, ex. 2dc55d50-3f6c-446a-87d0-a5a4eed23269</returns>
+        /// <remarks>
+        /// ```scriban-html
+        /// {{ math.uuid_v4 }}
+        /// ```
+        /// ```html
+        /// 1c0a4aa8-680e-4bd6-95e9-cdbec45ef057
+        /// ```
+        /// </remarks>
+        public static string UuidV4()
```
xoofx, line 353: You can expose `uuid` instead of `uuid_v4`. The API is mostly user oriented that don't care whether it is v4 or v1.

src/Scriban/Functions/MathFunctions.cs
```diff
@@ -25,6 +25,8 @@ namespace Scriban.Functions
 #endif
     partial class MathFunctions : ScriptObject
     {
+        private static readonly Random _random = new Random();
```
xoofx, line 28: A static object is not great here, because Scriban can be used from multiple threads. You can store the `Ramdom` in `TemplateContext.Tags`  instead.

src/Scriban/Functions/MathFunctions.cs
CommonGuy: Ok, changed this to `uuid`.

src/Scriban/Functions/MathFunctions.cs
CommonGuy: Ups, didn't account for multi threading. Fixed.

xoofx:
Thanks!

PR #343, opened 2021-05-02
add complementary operator to null-coalescing operator (??) -> (?!)

NeVeSpl:
In Scriban operators <code>||</code> and <code>&&</code> always return <code>boolean</code>. They behave as in normal strongly typed languages. But when working with string templates in jsx/react I found that JS like behaviour is very useful. Right now Scriban has equivalent of JS <code>||</code> in form of <code>??</code>. What I am proposing is adding complementary operator <code>?!</code> that would work as JS <code>&&</code>. 

```
a ?? b
where is defined as:
a != null ? a : b;
```

```
a ?! b
where is defined as:
a != null ? b : a;
```


xoofx:
Thanks!

PR #344, opened 2021-05-02
fix not detecting function call after pipe call when function does not have any arguments

NeVeSpl:
For the given template, when a function has some argument we get the correct AST:
```
{{ var | func 2 }} 
```
```
ScriptPage (0 - 20)
  ScriptBlockStatement (0 - 20)
    ScriptEscapeStatement (0 - 1)  [{{]
    ScriptExpressionStatement (3 - 14)
      ScriptPipeCall (3 - 14)
        ScriptVariableGlobal (3 - 5)  [var]
        ScriptToken (7 - 7)  [|]
        ScriptFunctionCall (9 - 14)
          ScriptVariableGlobal (9 - 12)  [func]
          ScriptLiteral (14 - 14)  [2]
    ScriptEscapeStatement (16 - 17)  [}}]
    ScriptRawStatement (18 - 20)
  ScriptBlockStatement END
```
but when a function does not have any arguments it is recognized as a global variable instead of as a function. 

```
{{ var | func }} 
```
```
ScriptPage (0 - 18)
  ScriptBlockStatement (0 - 18)
    ScriptEscapeStatement (0 - 1)  [{{]
    ScriptExpressionStatement (3 - 12)
      ScriptPipeCall (3 - 12)
        ScriptVariableGlobal (3 - 5)  [var]
        ScriptToken (7 - 7)  [|]
        ScriptVariableGlobal (9 - 12)  [func]
    ScriptEscapeStatement (14 - 15)  [}}]
    ScriptRawStatement (16 - 18)
  ScriptBlockStatement END
```

 Since after pipe operator, only valid node can be o type ScriptFunctionCall, this PR fixes that:

```
ScriptPage (0 - 18)
  ScriptBlockStatement (0 - 18)
    ScriptEscapeStatement (0 - 1)  [{{]
    ScriptExpressionStatement (3 - 12)
      ScriptPipeCall (3 - 12)
        ScriptVariableGlobal (3 - 5)  [var]
        ScriptToken (7 - 7)  [|]
        ScriptFunctionCall (9 - 12)
          ScriptVariableGlobal (9 - 12)  [func]
    ScriptEscapeStatement (14 - 15)  [}}]
    ScriptRawStatement (16 - 18)
  ScriptBlockStatement END
```

### DIFF
--- a/src/Scriban/Syntax/Expressions/ScriptPipeCall.cs
+++ b/src/Scriban/Syntax/Expressions/ScriptPipeCall.cs
@@ -69,20 +69,13 @@ namespace Scriban.Syntax
 
                 return result;
             }
-            catch (ScriptRuntimeException sre) when (sre.OriginalMessage.StartsWith("Invalid target function"))
-            {
-                newPipe = false; 
-                context.ClearPipeArguments();
-                throw new ScriptRuntimeException(To.Span, $"Pipe expression destination `{To}` is not a valid function ");
-            }
             catch
             {
                 // If we have an exception clear all the pipe froms
                 newPipe = false; // Don't try to clear the pipe
-                context.ClearPipeArguments();                
+                context.ClearPipeArguments();
                 throw;
             }
-          
             finally
             {
                 if (newPipe)
--- a/src/Scriban/Syntax/Expressions/ScriptPipeCall.cs
+++ b/src/Scriban/Syntax/Expressions/ScriptPipeCall.cs
@@ -69,13 +69,20 @@ namespace Scriban.Syntax
 
                 return result;
             }
+            catch (ScriptRuntimeException sre) when (sre.OriginalMessage.StartsWith("Invalid target function"))
+            {
+                newPipe = false; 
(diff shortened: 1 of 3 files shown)


src/Scriban/Syntax/Expressions/ScriptPipeCall.cs
```diff
@@ -69,13 +69,20 @@ public override object Evaluate(TemplateContext context)
 
                 return result;
             }
+            catch (ScriptRuntimeException sre) when (sre.OriginalMessage.StartsWith("Invalid target function"))
```
xoofx, line 72: It is too brittle to rely on exception message. I would prefer that in `ScriptFunctionCall` we directly detect the pipe and throw an exception with the correct message

You can move the part where you have this test:

```c#
var isPipeCall = processPipeArguments && context.CurrentPipeArguments != null && context.CurrentPipeArguments.Count > 0
```
And use this variable later in the function.

xoofx:
Thanks!

PR #346, opened 2021-05-07
333 - Add support for native indexers

binarycow:


src/Scriban/Syntax/Expressions/ScriptIndexerExpression.cs
```diff
@@ -130,17 +130,22 @@ private object GetOrSetValue(TemplateContext context, object valueToSet, bool se
             {
                 var accessor = context.GetMemberAccessor(targetObject);
                 var indexAsString = context.ObjectToString(index);
+                var itemAccessor = accessor as IItemAccessor;
 
                 if (setter)
                 {
-                    if (!accessor.TrySetValue(context, Index.Span, targetObject, indexAsString, valueToSet))
+                    var itemSuccess = itemAccessor?.ItemType == index?.GetType() && itemAccessor.TrySetItem(context, Index.Span, targetObject, index, valueToSet);
```
xoofx, line 137: Should probably use `context.ToObject` to `itemAccessor?.ItemType` instead to convert the index only if ItemType is not null and use the item accessor only in this case.

src/Scriban/Runtime/IItemAccessor.cs
```diff
@@ -0,0 +1,41 @@
...
+// Licensed under the BSD-Clause 2 license.
+// See license.txt file in the project root for full license information.
+#nullable disable
+
+using System;
+using Scriban.Parsing;
+
+namespace Scriban.Runtime
+{
+    public interface IItemAccessor
```
xoofx, line 11: Maybe add a comment to explain that it is for accessing indexers on object.

src/Scriban/Syntax/Expressions/ScriptIndexerExpression.cs
xoofx: (a source indexer value could be e.g long while the indexer is int, automatic casting is usually done in scriban for such case)

binarycow: Updated

src/Scriban/Runtime/IItemAccessor.cs
binarycow: Updated

xoofx:
LGTM, could you add a test for this support? Thanks!

xoofx:
It seems that some existing tests are failing.

binarycow:
Yep, workin on 'em.  Gotta prioritize my day job and all!

binarycow:
@xoofx - Try it now; tests are passing on my local machine.

xoofx:
Thanks!

Issue #345, opened 2021-05-06
Millisecond formatting using date.to_string with the %L specifier drops the number when it is all zeros


ChrisWiitamaki:
The %L specifier on [date.to_string](https://github.com/scriban/scriban/blob/master/doc/builtins.md#dateto_string) is documented to return the 3-digit millisecond value.  The corresponding [ruby documentation](https://apidock.com/ruby/DateTime/strftime#864-Complete-Formatting-Codes) lists the range as 000-999.

When the ms value is 0, however, it is dropped from the output.  I believe this is because DateTimeFunctions.cs is using the .NET [FFF](https://docs.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings#FFF_Specifier) format specifier instead of [fff](https://docs.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings#fffSpecifier):
     `{ 'L', (dateTime, cultureInfo) => dateTime.ToString("FFF", cultureInfo) }`

Here is test model, template, and the output:
_Model_
```
    {
        "nonZeroMillis": "May 6, 2021 10:04:35.123",
        "zeroMillis": "May 6, 2021 10:04:35.000"
    }
```
_Template_
```
    Non-zero millis:
        >{{ model.nonZeroMillis | date.parse | date.to_string "%L" }}<

    Zero millis (the zero characters are not in the output):
        >{{ model.zeroMillis | date.parse | date.to_string "%L" }}<
```
_Output_
```
    Non-zero millis:
        >123<

    Zero millis (the zero characters are not in the output):
        ><
```

xoofx:
Fixed in 3.8.0

PR #349, opened 2021-05-13
Modified behavior of %L formatter on date.to_string to not drop ms from time

ChrisWiitamaki:
Modified behavior of %L formatter on date.to_string to not drop milliseconds from the output when the value is zero.  (Issue #345)


xoofx:
Thanks!

PR #358, opened 2021-05-19
Fix compilation and typos (subtract)

szyb:
This is actually continue of PR #354
Also run code generations to regenerate *.generated.cs files and Docs

### DIFF
--- a/src/Scriban/ScribanAsync.generated.cs
+++ b/src/Scriban/ScribanAsync.generated.cs
@@ -856,17 +856,16 @@ namespace Scriban.Syntax
             var op = this.EqualToken.TokenType switch
             {
                 TokenType.PlusEqual => ScriptBinaryOperator.Add,
-                TokenType.MinusEqual => ScriptBinaryOperator.Substract,
+                TokenType.MinusEqual => ScriptBinaryOperator.Subtract,
                 TokenType.AsteriskEqual => ScriptBinaryOperator.Multiply,
                 TokenType.DivideEqual => ScriptBinaryOperator.Divide,
                 TokenType.DoubleDivideEqual => ScriptBinaryOperator.DivideRound,
                 TokenType.PercentEqual => ScriptBinaryOperator.Modulus,
-                _ => throw new NotImplementedException()
+                _ => throw new ScriptRuntimeException(context.CurrentSpan, $"Operator {this.EqualToken} is not a valid compound assignment operator"),
             }
 
             ;
-            var returnValue = ScriptBinaryExpression.Evaluate(context, this.Span, op, left, right);
-            return returnValue;
+            return ScriptBinaryExpression.Evaluate(context, this.Span, op, left, right);
         }
     }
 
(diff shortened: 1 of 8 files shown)


xoofx:
Thanks! Could you rebase as it seems that you are embedding in this PR changes coming from another PR?

szyb:
No problem, however my base for this PR was latest master (which is failing to build) `548daf6 - Merge pull request #354 from binarycow/typo_in_subtract_operator` 
Should I rebase it to which commit?

src/Scriban/ScribanAsync.generated.cs
```diff
@@ -1289,9 +1288,17 @@ public static async ValueTask<object> CallAsync(TemplateContext context, ScriptN
 
             var scriptFunction = functionObject as ScriptFunction;
             var function = functionObject as IScriptCustomFunction;
+            var isPipeCall = processPipeArguments && context.CurrentPipeArguments != null && context.CurrentPipeArguments.Count > 0;
             if (function == null)
             {
-                throw new ScriptRuntimeException(callerContext.Span, $"Invalid target function `{functionObject}` ({context.GetTypeName(functionObject)})");
+                if ((isPipeCall) && (callerContext is ScriptFunctionCall funcCall))
```
xoofx, line 1294: These changes are coming from a different PR that were merged. So it doesn't look like you are on master.

szyb: Hmm... Actually I was on master. This code was generated (changed) when I run Scriban.AsyncCodeGen (together with line 859 in ScribanAsync.generated.cs). I assumed this code generator was not run lately, so it changed not only the line I wanted, but also other lines that were not up-to-date. Am I wrong?

xoofx: Oh, ok, maybe that's the issue, the original PR forgot to make the `AsyncCodeGen` and that was missed
FYI @binarycow

binarycow: Yeah... sorry @xoofx... I forgot about the code gen...

I'm curious tho... the build task in the original PR (#354) passed.  @szyb says that `master` was failing to build.  Shouldn't that not be possible, if the build task passes before its merged?

Also, what are your thoughts about adding the codegen/docgen process to the CI/CD pipeline?

szyb: @binarycow Failing build after merge was because other PR #353 (which was later merged to the master than your base commit). It added method `GetValueToSet` in `StringAssignExpression` which uses `ScriptBinaryOperator.Substract` (with the typo) and PR #354 renamed this enum value. After merge it failed to build.

I think it might be a good idea to generate the code/docs in CI/CD.

xoofx: > I think it might be a good idea to generate the code/docs in CI/CD.

Yes, as a verification test. I don't want the CI to generate extra commits on top of users ones, but we do want to catch that if someone made a change and forgot to update/run the file, we can detect this with a unit test.

I need to add a message for PR contribution so that also folks making a PR don't forget these steps when they contribute. It's currently not written anywhere, so it's really easy to miss them - and for me to forget about them during PR reviews.

Issue #360, opened 2021-05-25
ternary operator precedence changed with 3.8.0?


SvenMitlehner:
I was trying version 3.8.0 and stumbled upon a changed behaviour compared to 3.7.0 when using the ternary operator. 

I could not find any information about it in the changelog, so i assume it might not be an intended change. (I am aware of https://github.com/scriban/scriban/issues/355, but it seems to be a different problem)

Evaluating the template

`{{x = '3'; '1' | string.contains '2' ? '' : x}}`

gives the error message

`Invalid target function 3`

When adding brackets to the condition it works as expected

`{{x = '3'; ('1' | string.contains '2') ? '' : x}}`

Not using the variable x also evaluates without an error

`{{x = '3'; '1' | string.contains '2' ? '' : '3'}}`

The faulty template worked fine in 3.7.0

xoofx:
Likely a regression not caught by tests. We will have a look (cc: @binarycow could have been introduced by some recent changes?)

NeVeSpl:
That my fault, caused by  #344, x is detected as a function call and should be as variable.

xoofx:
> That is my fault, caused by #344, x is detected as a function call and should be as variable.

Oh, ok, I'm going to revert it and publish a 3.8.1 in the meantime

xoofx:
The revert should be available in a few minutes in `3.8.1`

PR #361, opened 2021-05-25
fix not detecting function call after pipe call when function does not have any arguments, second try

NeVeSpl:
It restores #344 and also eliminates the reason why #360 appeared.


src/Scriban/Parsing/Parser.Expressions.cs
```diff
@@ -660,6 +660,15 @@ private ScriptExpression ParseExpression(ScriptNode parentNode, ScriptExpression
                         break;
                     }
 
+                    if ((!_isScientific) && (leftOperand is ScriptVariableGlobal) && (parentNode is ScriptPipeCall) && (functionCall == null))
```
xoofx, line 663: I don't understand this line related to the tests added above. If `leftOperand is ScriptVariableGlobal` is the case, why the tests above involves `string.contains '2'`, doesn't look like the tests are actually testing this fix, or am I missing something?

NeVeSpl: Test that I have added is exactly the same use case as reported in #360 that showed regression. This test makes sure that fix is no applied at all in this test scenario. Previously (in #344) this ``if`` was passing for the last ``x``, and ``x`` was incorrectly changed to ScriptFunctionCall. It was happening because  ``parentNode`` for last ``x`` was incorrectly set to ScriptPipeCall. So this test makes sure that the parent node for the last ``x`` is not of type ScriptPipeCall and fix is not applied.

xoofx: oh ok, I see the `parentNode` changes above.

But then, where is the test to verify that we don't regress on #360 ?
And to understand the original issue, this was about just the AST, the runtime was working correctly right?
Also wondering why restricting on `leftOperand is ScriptVariableGlobal`? (e.g imagine you have something like `1 | a.b` that would not be caught by this change as well?)

NeVeSpl: This is that test, it checks that regression reported in #360 will not happen again.

Yes the original issue is only about changing one node in AST in a very specific situation, when after pipe operator we have function call without any additional parameter.  Runtime works correctly regardless if AST for this case is fixed or not. 
I need to check that restriction later because it should be applied in that case ``1 | a.b``, it is the most important case for me, I have a lot of function invocation written exactly that way.

xoofx: > This is that test, it checks that regression reported in #360 will not happen again.

Sorry I was not clear, not talking about this test in particular. I was referring to a unit test (e.g checking the result of the AST)

> Yes the original issue is only about changing one node in AST in a very specific situation,

Ok, so in that case, that needs to be solved not for a specific situation. Sorry If I didn't pay enough attention to the original PR.

NeVeSpl: Are they any test for AST right now? I do not see any...

Indeed, the fix doesn't work for ``1 | a.b``, but it should, I Will try to come up with a more general solution.

xoofx: > Are they any test for AST right now? I do not see any...

Not any, you are the first for asking such a thing :)
You can add it next to some unit tests we already have.

NeVeSpl:
I wonder what is precedence between function call/pipe call and operators, for this example:
```
{{ ""123"" | string.to_int + 1 }}
{{ string.to_int ""123"" + 1 }}
```
runtime gives the following answers
```
124
1231
```
for my taste, they should be the same. From AST perspective first input should not render at all, but runtime does some magic and is able to pipe a value to binary expression:
```
==
ScriptPage (0 - 94)
  ScriptBlockStatement (0 - 94)
    ScriptEscapeStatement (0 - 1)  [{{]
    ScriptExpressionStatement (3 - 27)
      ScriptPipeCall (3 - 27)
        ScriptLiteral (3 - 7)  ["123"]
        ScriptToken (9 - 9)  [|]
        ScriptBinaryExpression (11 - 27)
          ScriptMemberExpression (11 - 23)
            ScriptVariableGlobal (11 - 16)  [string]
            ScriptToken (17 - 17)  [.]
            ScriptVariableGlobal (18 - 23)  [to_int]
          ScriptToken (25 - 25)  [+]
          ScriptLiteral (27 - 27)  [1]
    ScriptEscapeStatement (29 - 30)  [}}]
    ScriptRawStatement (31 - 61)
    ScriptEscapeStatement (62 - 63)  [{{]
    ScriptExpressionStatement (65 - 87)
      ScriptFunctionCall (65 - 87)
        ScriptMemberExpression (65 - 77)
          ScriptVariableGlobal (65 - 70)  [string]
          ScriptToken (71 - 71)  [.]
          ScriptVariableGlobal (72 - 77)  [to_int]
        ScriptBinaryExpression (79 - 87)
          ScriptLiteral (79 - 83)  ["123"]
          ScriptToken (85 - 85)  [+]
          ScriptLiteral (87 - 87)  [1]
    ScriptEscapeStatement (89 - 90)  [}}]
    ScriptRawStatement (91 - 94)
  ScriptBlockStatement END
==

```

xoofx:
Frankly, we can't really resolve this:

```
{{ ""123"" | string.to_int + 1 }}
{{ string.to_int ""123"" + 1 }}
```

You need to use parenthesis to get the correct behavior.

``
{{ ""123"" | string.to_int + 1 }}
{{ (string.to_int ""123"") + 1 }}
```
For this case:

``` 
{{ 1 | math.plus 2 * 2 }}
{{ math.plus 1 2 * 2 }}
```

This is normal according to precedence between binary operators and function/pipe calls from a parser/execution perspective, because this is equivalent to:

``` 
{{ 1 | math.plus (2 * 2) }}
```

**Comment by NeVeSpl** [comment]
But if operators have higher precedense should not:
```
{{ "123" | string.to_int + 1 }}
```
be equivalent to:
```
{{ "123" | (string.to_int + 1) }}
```
and throw exception?

**Comment by xoofx** [comment]
> and throw exception?

Yes, that's a particular case for nested parenthesis, they protect from flowing with pipe arguments:

https://github.com/scriban/scriban/blob/53da5bc08964319be948638f2747f8610b305e3c/src/Scriban/Syntax/Expressions/ScriptNestedExpression.cs#L74-L89

Which is not the case for binary operations, but I'm too worried that fixing it correctly would cause severe performance regressions.

If you have a doubt use parenthesis. 

In general, if you use operators behind a pipe call, you need to protect things correctly with parenthesis.
I will close this issue for now as won't fix.

**Comment by NeVeSpl** [comment]
Could you reopen it? This PR is about adding function call after pipe call, and not changing existing behaviour regarding pipe call and operators. I can live with that as it is, but I still need that function call in AST to appear.

**Comment by xoofx** [comment]
> Could you reopen it? This PR is about adding function call after pipe call, and not changing existing behaviour regarding pipe call and operators. I can live with that as it is, but I still need that function call in AST to appear.

I can but, I would like to understand: Why this is so important for you? What are you using from this? 🙂

**Comment by NeVeSpl** [comment]
I am using AST for syntax coloring, and functions should have a different color than variables. And as you can see it does not work well right now:

image

**Comment by xoofx** [comment]
Ok, thanks I see, makes sense.

**Comment by NeVeSpl** [comment]
I have added support for member expression `1 | a.b` and I am not seeing a way to make fix more general since AST is flawed and leftOperand can be a binary expression. This is probably the safest whey without introducing any braking change in the runtime.

In next commit I will introduce tests for AST.

File: src/Scriban/Parsing/Parser.Expressions.cs [hunk 2]
```diff
@@ -660,6 +660,18 @@ private ScriptExpression ParseExpression(ScriptNode parentNode, ScriptExpression
                         break;
                     }
 
+                    if ((!_isScientific) && (parentNode is ScriptPipeCall) // after a pipe call we expect to see a function call
+                                         && (functionCall == null)         // but when function is not followed by any parameter e.g. '1 | math.abs', above code does not create function call,
+                                                                           // here we fix that by creating function call, but only when leftOperand is e.g. '1 | abs' or '1 | math.abs' 
+                                         && (leftOperand is ScriptVariableGlobal || leftOperand is ScriptMemberExpression) // we need that restriction since leftOperand can be of other type e.g. binary expression '"123" | string.to_int + 1'
```
xoofx, line 666: You should check instead for `leftOperand is IScriptVariablePath` because we will be missing many other cases like `1 | $myvar` or `1 | myvar[1]`

xoofx, line 666:
```suggestion
                                         && (leftOperand is IScriptVariablePath) // we need that restriction since leftOperand can be of other type e.g. binary expression '"123" | string.to_int + 1'
```

NeVeSpl:
Thank you, now everything looks like it should 😄 

image

Issue #362, opened 2021-05-26
Regression: Error during runtime which did not occur with 3.7.0


badcel:
I get the following error if using Scriban. The error does not occur with version 3.7.0

> /home/runner/.nuget/packages/scriban/3.8.1/src/Scriban/Runtime/IItemAccessor.cs(38,14): error CS0051: Inconsistent accessibility: parameter type 'SourceSpan' is less accessible than method 'IItemAccessor.TrySetItem(TemplateContext, SourceSpan, object, object, object)' [/home/runner/work/gir.core/gir.core/Integration/Integration.csproj]

I think the newly introduced `ItemAccessor` has to respect the scriban compile time constants like in `SourceSpan`:

https://github.com/scriban/scriban/blob/162ad681e381505486fc7dd4bf201075b8d55c04/src/Scriban/Runtime/IItemAccessor.cs#L15

```cs
#if SCRIBAN_PUBLIC    
    public
#else    
    internal
#endif    
struct SourceSpan
```

We use the following setting: `<PackageScribanIncludeSource>true</PackageScribanIncludeSource>`

I did not have the time to create a sample which shows the problem but I think the error is quite clear.

The failed PR is here: https://github.com/gircore/gir.core/pull/335

If nobody steps up I can likely fix this in the coming days.

xoofx:
Good catch, we don't have a good testing story for this right now. cc:  @binarycow

binarycow:
[PR Submitted](https://github.com/scriban/scriban/pull/363)

Sorry about that!

xoofx:
Thanks @binarycow! A new version `3.8.2` is on its way...

PR #363, opened 2021-05-26
362 - Add conditional compilation for source including

binarycow:


binarycow:
Fixes #362

xoofx:
Thank you!

Issue #366, opened 2021-06-22
Lazy Load Values


Adam-Howarth:
We have a similar issue that has been reported in #323. We have created class that lazy load all properties but are running into the same problem that all of the properties are being evaluated despite not being used.

Adding a dummy class as the root object and referencing all properties using "model." (the example in #323) isn't really a viable solution for us. We have scoped out the changes for creating a PR that will allow us to keep the models as is through creating a "Lazy" script object. Which would also require changes to the extension method Import so set those values correctly.

But before we go down that route, has this approach been explored before? If so were there any problems that arose? Is there a workaround outside of moving the properties off of the route?

If not we are willing create a PR but wanted to gauge the likelihood of that being accepted. Naturally you would have to see the code first, but reaching out in case there is something we can do beforehand.

SRBobGilliland:
@xoofx Any thoughts?

xoofx:
> @xoofx Any thoughts?

I'm on holidays, I will try comment back in August

xoofx:
You can push your own top level `IScriptObject` to a template context (`TemplateContext.PushGlobal`), as long as it respects the contract, you should be able to implement any lazy loading you want, redirecting to any underlying managed objects or not. You would have also to respect the storing part of the contract (e.g store variable).
I would suggest to implement this on your side first, you can share the code here if you want. But not sure I would like to maintain such a scenario. I'm worried it can be very use case specific, some would start to come with other requirements (e.g I want to bind multiple properties from different objects on the top level script but with lazy...etc.)...
Anyway, you should not be blocked to implement such feature on top of Scriban without forking Scriban. Let me know.

ShiroMiyuki:
Hi, I tried implementing it on my end it worked somehow ok, but there is an issue with importing the object, as we can add our 
delegates to try to change the value we retrieved in the object, but we also need to wrap the values into lazy objects / or Func's and that logic is sitting inside the ScriptObjectExtensions:::Import method.

I forked the solution and the easiest way to go around it was indeed not to implement lazy loading inside the existing solution to satisfy our needs but to add the delegates and expose them in the import method.

I ran the tests and all of them passed as well as I ran the performance checks. The difference in code before and after code change seems to not impact the overall performance too much (it's -3 to 3 us less/more so I think it's within margin error)

I raised PR for the change
https://github.com/scriban/scriban/pull/376

ShiroMiyuki:
[...]

PR #376, opened 2021-08-17
Expose property and field setters in the import extension method

ShiroMiyuki:
Raising PR for the issue
https://github.com/scriban/scriban/issues/366


xoofx:
Closing as per comment [here](https://github.com/scriban/scriban/issues/366#issuecomment-903297984)

PR #380, opened 2021-09-13
Conversion between a non-nullable and nullable type can throw an exce…

JvanderStad:
I had an exception while converting an `Int32` to a `Nullable<Int64>`. The `TrySetValue` method now converts the prop/field value with method `ChangeType`.

`ChangeType` checks if the target type is a nullable and tries to change the type to the underlaying type of the nullable type

### DIFF
--- a/src/Scriban/Runtime/Accessors/TypedObjectAccessor.cs
+++ b/src/Scriban/Runtime/Accessors/TypedObjectAccessor.cs
@@ -95,37 +95,21 @@ namespace Scriban.Runtime.Accessors
 
         public bool HasIndexer => _indexer != null;
 
-        public static object ChangeType(object value, Type targetType)
-        {
-            if (value == null)
-                return null;
-
-            if (value.GetType() == targetType)
-                return value;
-
-            if (targetType.IsGenericType && targetType.GetGenericTypeDefinition() == typeof(Nullable<>))
-                targetType = Nullable.GetUnderlyingType(targetType);
-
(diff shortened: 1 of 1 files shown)


src/Scriban/Runtime/Accessors/TypedObjectAccessor.cs
```diff
@@ -95,20 +95,32 @@ public bool TrySetItem(TemplateContext context, SourceSpan span, object target,
...
-            MemberInfo memberAccessor;
-            if (_members.TryGetValue(member, out memberAccessor))
+            if (_members.TryGetValue(member, out MemberInfo memberAccessor))
             {
-                var fieldAccessor = memberAccessor as FieldInfo;
-                if (fieldAccessor != null)
+                if (memberAccessor is FieldInfo fieldAccessor)
                 {
-                    fieldAccessor.SetValue(target, value);
+                    fieldAccessor.SetValue(target, ChangeType(value, fieldAccessor.FieldType));
```
xoofx, line 118: Conversion between value to a new type should use `context.ToObject` instead. `ChangeType` should not be introduced

JvanderStad: Righto. Updated

JvanderStad:
@xoofx Any chance you could merge this PR? Thanks! 👍

xoofx:
Yep, sorry, thanks!

PR #384, opened 2021-10-08
Support for specifying object parameter type when importing delegates…

JvanderStad:
Hi Alexandre,

I have added support for specifying object parameter type when importing delegates.

🖐 Cheers


JvanderStad:
I missed the fact `ToObject` was a `virtual`. I can do what I want in the `override`. Thanks :thumbsup:

PR #400, opened 2022-01-06
fix typo - `endfor` is used for liquid, which actually would like to avoid in scriban language.

imzjy:


imzjy:
ref: http://xoofx.com/blog/2017/11/13/implementing-a-text-templating-language-and-engine-for-dotnet/#control-flows

xoofx:
Yep good catch, thanks!

PR #401, opened 2022-01-09
fix accessing null?.property when EnableRelaxedMemberAccess == False

NeVeSpl:
When `TemplateContext.EnableRelaxedMemberAccess` is set to `false` this template fails:
```
{{
x = null     # x is null
x?.y          # x?.y will fail
}}
```
but it should not, since we use here `?.` and not `.`


xoofx:
Good catch, thanks!

Issue #392, opened 2021-11-23
Grammar for `when` not allowing a negative number


mikechristiansenvae:
(In this example, `myDate` is a `System.DateTimeOffset`)
```
case myDate.offset.hours
    when -5
        " EST"
    when -4
        " EDT"
    else
        " ???"
end
```

Produces:

```
When is expecting at least one value.
Invalid token found `-`. Expecting <EOL>/end of line.
```

Changing the `-5` and `-4` to `5` and `4` results in the template compiling fine.

xoofx:
This is fixed

PR #407, opened 2022-01-29
Add support for negative numbers in 'when' statements

AlfredoMS:
Fix for issue #392 

This change adds support for negative numbers, both integers and float, to be added as conditions for "when" statements.

Let me know if I should add more tests.


lofcz:
This should probably not be limited to `when -10` but also support cases like `when +-+-10`.

AlfredoMS:
hey @lofcz thanks for the feedback, however I think that would be of a much bigger scope than this change intends. I think bringing functionality parity of literals and constants with negative numbers makes sense and has immediate benefits,

I am thinking adding support for full constant expression might be a good follow up change. How does that sound? Otherwise I'll start taking a look into what would it take to accomplish that, maybe if it is a short change I'll just update this PR

lofcz:
Seems like #409 superseeds this. I would recommend to close this PR and continue discussion there.

xoofx:
Fixed by #409 
Thanks!

PR #409, opened 2022-02-01
Add support for constant expressions conditions in case-when statements

AlfredoMS:
As a result of some feedback on PR #407 
This is my first try to add full support for constant expressions on case-when statements.

Constant expressions are those which could be fully evaluated at compile time, so it basically excludes increment/decrement operators and function calls.

Probably need more test cases or some suggestions on things I might be missing.

Still need to test functions calls, which should be disallowed.


lofcz:
The supported scope is now a bit broader than I've had on mind originally. Expressions with variables are not constant expressions (we can't fold this into a single literal before running the ast tree).

Try to paste this into your C# editor:
```cs
int a = 1;
switch (10)
{
 case 2 + 2: // ok, can be folded into 4
 {
                    
 }

 case a: // err; we don't know value of a at compile time here
 {
                    
 }
}
```

Current situation is a bit unfortunate as for example when `case-when` would be used within a loop and one of `when` expression would be `2+2` then this expression would need to be interpreted in every iteration (negating benefit of said constant expression constraint). 

On the other hand allowing expressions in general here might be a better fit for a dynamic lang. such as Scriban. 

We might also try to fold what can be folded in a separate pass but still allow for non-constant expressions to be used. This could be done as a follow up PR.

```js
case test_value
  when 2+2 // <-- we can fold this
  when myVar + 2 // <-- this has to be evaluated ad-hoc
end
```

I would be interested in @xoofx's thoughts about this.

xoofx:
Thanks for trying to bring support.

I will have a deeper look at this PR locally because I don't recall the details of the code impacted.

One thing is that you can probably let any kind of expressions (including `x++`) because any variables in scriban could lead potentially to a function call that could have side effects. So trying to restrict damage in one place is not enough if we wanted to do it consistently.

It's probably fine to support any expressions.

AlfredoMS:
@lofcz @xoofx , thanks for the comments.

I was also afraid of the broad scope, hence wanted to limit my initial contribution to negative numbers as per PR #407 . (Which I still suggest to merge first in case this one takes too long). My initial thought was also that we should try to limit the scope of what is allowed, so could be ok to restrict the use of variables and function calls. At the same time in my tests it looks like using variables works fine, as long as we don't try to modify them at the same time we evaluate them.

One question I have related to the failing Liquid test is about the meaning of `or` and `||`. Based on the code comments for the case-when statement in Liquid `or` should be treated as `,`; and same for the double pipe in Scriban. This obviously conflicts with the `boolean or` operator.

On a quick check to Liquid docs (https://shopify.github.io/liquid/tags/control-flow/) and Scriban docs (https://github.com/scriban/scriban/blob/master/doc/language.md#94-case-and-when). There is no mention of such uses of `or` and `||` as list separators. So is it ok as a breaking change to only use `,` to set multiple conditions?

xoofx:
Thanks! I have merged your PR manually and slightly change the code so that it accepts any expression (the protection against ++/-- is not enough anyway so...)
I have fixed also the issue with the `||` that will stop parsing if we are parsing a when expression

PR #410, opened 2022-02-05
Add net6.0

tibel:
Added .NET 6.0 and updated package references.

Also fixed an `AmbiguousMatchException` with `Queryable.Take` that occurs on .NET 6.0.


xoofx:
Thanks, gonna check that latest NuDoq integrated the fix I had to do.

PR #415, opened 2022-02-10
413: string.index_of

binarycow:


xoofx:
Merged by commit 72f80e8

Issue #420, opened 2022-03-09
Proposal/runtime - treat last param as being declared with "params" modifier if of type "object[]" to support local functions


lofcz:
Consider this snippet:
```cs
delegate string Args(object[] args);

public class Cls2
{
    public string str = "hi";
}

[TestFixture]
public class TestRuntime
{
    string Dump(params object[] args)
    {
        return "hello";
    }

    [Test]
    public void TestPars()
    {
        ScriptObject model = new ScriptObject();
        ScriptObject debug = new ScriptObject();
        Args dump = Dump;
        debug.Import("dump", dump);
        
        model["debug"] = debug;
        model["native"] = new Cls2();
        
        var input = "{{debug.dump(native, native)}}";
        var template = Template.Parse(input);
        var result = template.Render(model);
        Assert.AreEqual("hello", result);
    }
}
```

This test passes. Now if `Dump` is defined inside `TestPars` instead:

```cs
[Test]
public void TestPars()
{
    string Dump(params object[] args)
    {
        return "hello";
    }

    ScriptObject model = new ScriptObject();
    ...
}
```

Compiler is still happy about this. Once script runs `Runtime/DynamicCustomFunction` resolves cardinality of `Dump` in ctor. To determine whether `args` is marked with `params` we are looking for `param.GetCustomAttributes(typeof(ParamArrayAttribute))` which we get in the first case:

![image](https://user-images.githubusercontent.com/10260230/157375344-6f7fc8db-f2d9-48da-b94a-060baeb2a8c8.png)
 
But in the other case `Dump` is compiled to `g__Dump` and the attribute lost.

![image](https://user-images.githubusercontent.com/10260230/157375581-babb2aa1-3b0e-42d4-ae4b-a38b75f96f40.png)
![image](https://user-images.githubusercontent.com/10260230/157375609-d73ed6f3-fc28-4ed3-aeba-dd8734a341e4.png)

As a result `_paramsIndex` is still `-1` and we either get an error about cardinality mismatch or "unable to convert type A to B" from `TemplateContext.Helpers`.

This could result in confustion for end users as it's unclear why the first case works and the other does not. There are more ways to solve this but couldn't we just assume that any method declared with `object[]` as last param is intended as `params object[] param`? This would render resolution trivial and we would also avoid a call to `lastParam.GetCustomAttributes`.

xoofx:
Agreed, that should be fine. PR welcome.

lofcz:
resolved in #421

PR #421, opened 2022-03-09
Support local methods with params modifier

lofcz:
Implementation of #420 🪴

### DIFF
--- a/src/Scriban/Runtime/DynamicCustomFunction.cs
+++ b/src/Scriban/Runtime/DynamicCustomFunction.cs
@@ -75,12 +75,21 @@ namespace Scriban.Runtime
                 var lastParam = Parameters[Parameters.Length - 1];
                 if (lastParam.ParameterType.IsArray)
                 {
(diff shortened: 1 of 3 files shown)


src/Scriban/Runtime/DynamicCustomFunction.cs
```diff
@@ -75,12 +75,21 @@ protected DynamicCustomFunction(MethodInfo method)
                 var lastParam = Parameters[Parameters.Length - 1];
                 if (lastParam.ParameterType.IsArray)
                 {
-                    foreach (var param in lastParam.GetCustomAttributes(typeof(ParamArrayAttribute), false))
+                    if (lastParam.ParameterType == typeof(object[]))
```
xoofx, line 78:
```suggestion
                    if (lastParam.ParameterType == typeof(object[]) || lastParam.ParameterType == typeof(ScriptExpression[]))
```

Adding support for `ScriptExpression[]` as well.

xoofx:
Thanks!

PR #426, opened 2022-04-03
Fix enum array contains

### LINKED ISSUE #425: array.contains doesn't work with arrays of enums
Given an object like:

```csharp
public class WeatherForecastSearchRequest
{
    [FromQuery(Name = "strings")]
    public List<string> Strings { get; set; } = new List<string>();

    [FromQuery(Name = "since")]
    public DateTime? Since { get; set; }

    [FromQuery(Name = "types")]
    public List<WeatherType> Types { get; set; } = new List<WeatherType>();

    [FromQuery(Name = "verbose")]
    public bool Verbose { get; set; }
}

public enum WeatherType
{
    Guess,
    Known
}
```

And a template like:

```html
<div class='search'>
    <form name='Search' method='get' action>
        <div class='property'>
            <label class='name' for='strings'>Strings:</label>
            <input id='strings' type='text' name='strings' value='{{ Form.Strings | array.join ',' }}' />
        </div>
        <div class='property'>
            <label class='name' for='since'>Since:</label>
            <input id='since' type='datetime-local' name='since' value='{{ Form.Since | date.to_string '%Y-%m-%dT%H:%M:%S' }}' />
        </div>
        <div class='property'>
            <label class='name' for='types'>Types:</label>
            <select name='types' id='types' multiple>
                <option value='Guess' {{ selected = ''; if Form.Types | array.contains 'Guess'; selected = 'selected'; end; selected }}>Guess</option> // this line doesn't work as is
                <option value='Known' {{ selected = ''; if Form.Types | array.contains 'Known'; selected = 'selected'; end; selected }}>Known</option> // this line doesn't work as is
            </select>
        </div>
        <div class='property'>
            <label class='name' for='verbose'>Verbose:</label>
            <input id='verbose' type='checkbox' name='verbose' value='true' {{ checked = ''; if Form.Verbose; checked = 'checked'; end; checked }}/>
        </div>
        <input type='submit' value='Search'>
    </form>
</div>
```

Lines like the following do not work as expected.

```html
<option value='Guess' {{ selected = ''; if Form.Types | array.contains 'Guess'; selected = 'selected'; end; selected }}>Guess</option>
```

Instead we have to do the following to marshall the array to an array of strings.
```html
<option value='Guess' {{ selected = ''; if Form.Types | array.each @string.strip | array.contains 'Guess'; selected = 'selected'; end; selected }}>Guess</option>
```

cspital:
Resolves #425

### DIFF
--- a/src/Scriban.Tests/TestFunctions.cs
+++ b/src/Scriban.Tests/TestFunctions.cs
@@ -45,6 +45,7 @@ namespace Scriban.Tests
                 TestParser.AssertTemplate("true", "{{ value | array.contains 100 }}", model: new ObjectModel { Value = mixed });
                 TestParser.AssertTemplate("false", "{{ value | array.contains 'Second' }}", model: new ObjectModel { Value = mixed });
                 TestParser.AssertTemplate("false", "{{ value | array.contains 101 }}", model: new ObjectModel { Value = mixed });
+                TestParser.AssertTemplate("false", "{{ value | array.contains 'Third' }}", model: new ObjectModel { Value = mixed });
             }
             class ObjectModel
             {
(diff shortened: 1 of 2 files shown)


xoofx:
Thanks!

PR #430, opened 2022-04-28
Add extensions that extract FrontMatter as a Dictionary<string, string>

jaguire:
I had a need to access the front matter from a template so I wrote an extension. 😄

```text
+++
variable = 1
name = 'yes'
+++
TEMPLATE
```
```csharp
var options = new LexerOptions { Mode = ScriptMode.FrontMatterAndContent };
var template = Template.Parse(source, lexerOptions: options);
var values = template.Page.FrontMatter.GetValues(); // returns a Dictionary<string, string>
```

I added unit tests too. ✨

### DIFF
--- a/src/Scriban.Tests/TestParser.cs
+++ b/src/Scriban.Tests/TestParser.cs
@@ -417,8 +417,53 @@ variable + 1
             TextAssert.AreEqual("This is after the frontmatter: yes\n2", pageResult);
         }
 
+        [Test]
+        public void TestFrontMatterExtensions_WithAssignExpression()
+        {
+            var options = new LexerOptions { Mode = ScriptMode.FrontMatterAndContent };
+            var input = @"+++
+variable = 1
+name = 'yes'
++++
+TEMPLATE";
+            input = input.Replace("\r\n", "\n");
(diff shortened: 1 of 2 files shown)


xoofx:
Thanks, but unfortunately, I will have to decline this PR. It's ok to keep this extension method in your project, if it fits your usage.

Extracting values directly from the syntax tree is a too narrow use case, because even in a Scriban front matter, you can perform calculations, call functions...etc.

So the proper way to do it is to evaluate the frontmatter (as for any syntax construct) with a TemplateContext. There is an example [here](https://github.com/lunet-io/lunet/blob/b34ad38e3981f938ecc4b5b11ccbf0834e744e08/src/Lunet.Core/Core/ScriptingPlugin.cs#L198-L231) and you can see that depending on your context use case, it can be more involving.

jaguire:
I did look at `template.Page.FrontMatter.Evaluate(context)`. It returns a null so I just assumed I was accessing some internal method that wasn't applicable at that time. Apparently, I didn't look close enough or I would have seen that context.CurrentGlobal gets filled with front matter after the call to evaluate(). 🤣 This is exactly what I needed. 👍🏽

PR #457, opened 2022-08-04
Fix a minor syntax error in the docs for case/when statements

bradtglass:
Fixes an (extremely small) syntax error in the language docs for case/when statements that caused me to go around in circles for quite a long time before I spotted it.


xoofx:
Oops, good catch, sorry for the trouble!

Issue #463, opened 2022-10-19
Unmatched end statement (at block level 1) terminates parsing


ChrisWiitamaki:
I recently ran into some behavior I wasn't expecting:  I accidentally had a extra {{end}} statement in my template.  The result was that parsing stopped at that point.  In effect, it acted as if it were the end of the template.

I looked in the documentation for this -- and perhaps a way to enforce strict matching of 'end to a block begin' -- but couldn't find anything.  When thinking about how this script-ending end statement may be used, it is logically problematic, as it can only be used when not inside another block.  In other words, it can only be used at the outermost level.

In order to prevent possible template bugs created by users (which was my situation), I would like to introduce a new ParserOption called EnforceEndStatementMustMatchBlockBegin.  The setting would default to false, thus ensuring that existing functionality is backwards compatible.  When setting this to true, if an end statement is found during parsing, and the _blockLevel is 1, an error will be logged.

The reason I went with a ParserOption vs. an option on the TemplateContext, is because the parser already logs an error if an end statement is not found for a block.  This setting just triggers the parser to enforce a match in the other direction.

I am attaching a pull request -- documented code with unit tests -- to aid in this request.  I'm open to suggestions for changes as well.  Thanks.

xoofx:
Thanks for the details!

I would be maybe more in favor to generate an error without the opt-in. I don't think there is an expectation about it and it is more error-prone as you experienced it.

xoofx:
(Rationale also is that the `return` statement is actually providing this behavior)

ChrisWiitamaki:
That sounds good.  I committed changes that remove the new parser option and simply generate an error on the condition.

ChrisWiitamaki:
Thanks for including this in the recent release!  I will close this issue.

Recording the error message that will be received in case it helps others stumble across the version containing the fix:
    "Found <end> statement without a corresponding beginning of a block"

PR #464, opened 2022-10-19
Implemented parsing option to enforce that an end-statement must match a block begin

ChrisWiitamaki:
Adds a new ParserOption to log an error and end-statement occurs on blockLevel 1.  Related to issue #463.

### DIFF
--- a/src/Scriban/Parsing/Parser.Statements.cs
+++ b/src/Scriban/Parsing/Parser.Statements.cs
@@ -52,6 +52,18 @@ namespace Scriban.Parsing
 
                     if (hasEnd)
                     {
+                        if (_blockLevel == 1 && Options.EnforceEndStatementMustMatchBlockBegin)
+                        {
+                            if (_isLiquid)
+                            {
+                                var syntax = ScriptSyntaxAttribute.Get(parentStatement);
+                                LogError(parentStatement, parentStatement?.Span ?? CurrentSpan, $"Found <end> statement `end{syntax.TypeName}` without a corresponding beginning of a block");
+                            }
+                            else
+                            {
+                                LogError(parentStatement, GetSpanForToken(Previous), $"Found <end> statement without a corresponding beginning of a block");
+                            }
+                        }
                         break;
                     }
                 }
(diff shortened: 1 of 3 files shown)


src/Scriban/Parsing/ParserOptions.cs
```diff
@@ -29,5 +29,11 @@ struct ParserOptions
         /// If the number cannot be represented to a decimal, it will fall back to a double.
         /// </summary>
         public bool ParseFloatAsDecimal { get; set; }
+
+        /// <summary>
+        /// <c>true</c> to return a parse error if an end statement does not correspond to a block (e.g. if-end, while-end, etc.).
+        /// Default is <c>false</c>: an end statement that is not part of a block is allowed, but will prematurely end the parsing.
+        /// </summary>
+        public bool EnforceEndStatementMustMatchBlockBegin { get; set; }
```
xoofx, line 37: As mentioned, we should probably remove the opt-in entirely.

ChrisWiitamaki: Removed.

src/Scriban/Parsing/Parser.Statements.cs
```diff
@@ -52,6 +52,18 @@ private ScriptBlockStatement ParseBlockStatement(ScriptNode parentStatement, boo
 
                     if (hasEnd)
                     {
+                        if (_blockLevel == 1)
+                        {
+                            if (_isLiquid)
+                            {
+                                var syntax = ScriptSyntaxAttribute.Get(parentStatement);
+                                LogError(parentStatement, parentStatement?.Span ?? CurrentSpan, $"Found <end> statement `end{syntax.TypeName}` without a corresponding beginning of a block");
```
xoofx, line 60: The error location seems to be off above. The log should probably use the `statement` instead of `parentStatement` (for both Scriban and Liquid) to get the correct Span location. In the case of an `end`, the statement should be the end statement.

xoofx: Fixed by 7804a97

ChrisWiitamaki:
Thanks for including this!

Issue #482, opened 2023-02-01
Extra space when an indented code block follows an {{~ if ~}} block


andradf:
I'm using this greatly appreciated library and seem to have run into a small issue with indentations. I am using version 5.5.2.
This template 
```
a
{{~ if true ~}}
b
{{~ end ~}}
 {{'c'}}
```
renders as 
```
a
b
  c
```
I believe we should expect only a single space indenting the c character. If we remove the first line, the if condition, or even declare c as a text block instead of a code block it renders correctly. For anyone else running into this, an easy workaround is to use a {{\u0020}} instead of the space.

andradf:
@xoofx, I looked at Scriban's code a bit and I think the problem might be in the Write method of TemplateContext.cs

```
public TemplateContext Write(string text, int startIndex, int count)
    {
        if (text != null)
        {
            // Indented text
            if (CurrentIndent != null)
            {
                var index = startIndex;
                var indexEnd = startIndex + count;

                while (index < indexEnd)
                {
                    // Write indents if necessary
                    if (_previousTextWasNewLine)
                    {
                        Output.Write(CurrentIndent, 0, CurrentIndent.Length);
                        _previousTextWasNewLine = false;
                    }

                    var newLineIndex = text.IndexOf('\n', index);
                    if (newLineIndex < 0 || newLineIndex >= indexEnd)
                    {
                        Output.Write(text, index, indexEnd - index);
                        break;
                    }

                    // We output the new line
                    Output.Write(text, index, newLineIndex - index + 1);
                    index = newLineIndex + 1;
                    _previousTextWasNewLine = true;
                }
            }
            else
            {
                Output.Write(text, startIndex, count);
            }
        }

        return this;
    }
```

If I change the if condition to `!string.IsNullOrEmpty(CurrentIndex)` it works as expected and we don't get a double space.

andradf:
I had a look at Scriban's code and I think I have identified the issue.  when comparing the cases 
```
a
b
 {{'c'}}
```
and 
```
a
{{~ if true ~}}
b
{{~ end ~}}
 {{'c'}}
```
[...]

PR #483, opened 2023-02-01
fix auto indent bug after ~}} block, add test

andradf:
Fixes issue #482 by resetting _previousTextWasNewLine to false when writing unindendented text.


xoofx:
Thanks!

PR #489, opened 2023-02-24
[Fix] Examples to 'limit, offster and reversed'

artur-simon:
Fix on the documentation as the actual format seem to be a pipe leading to the array functions.

Example 1 (not working): 

```
{{ 
    for i in 1..5 reversed
        i
    end 
}}
```

This gives us: `error : Invalid target function '5' (int)`

Example 2 (what actually worked):
```
{{ 
    for i in 1..5 | array.reversed
        i
    end 
}}
```

This gives us: `5 4 3 2 1` as expected.

This is the same for the `limit` and the `offset` parameters.

Tried it over the playground and it checks out.

https://scribanonline.azurewebsites.net/

### DIFF
--- a/doc/language.md
+++ b/doc/language.md
@@ -1151,7 +1151,7 @@ Allows to start the iteration of the loop at the specified zero-based index:
 
 > **input**
 ```scriban-html
-{{~ for $i in 4..9 offset:2 ~}}
+{{~ for $i in 4..9 | array.offset:2 ~}}
  {{ $i }}
 {{~ end ~}}
 ```
@@ -1169,7 +1169,7 @@ Limits the iteration of the loop to a specified count:
 
 > **input**
 ```scriban-html
-{{~ for $i in 4..9 limit:2 ~}}
+{{~ for $i in 4..9 | array.limit:2 ~}}
  {{ $i }}
 {{~ end ~}}
 ```
@@ -1185,7 +1185,7 @@ Reverses the iteration of the elements:
 
 > **input**
 ```scriban-html
-{{~ for $i in 1..3 reversed ~}}
+{{~ for $i in 1..3 | array.reversed ~}}
  {{ $i }}
 {{~ end ~}}
 ```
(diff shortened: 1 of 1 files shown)


xoofx:
It might be more a parsing bug than a documentation bug, because this should be supported (in the [test files](https://github.com/scriban/scriban/tree/master/src/Scriban.Tests/TestFiles/200-statements) there are offset/reversed test, but only with a simple variable for the iterator)

xoofx:
Actually It's not parsing related but [here](https://github.com/scriban/scriban/blob/85b2aabe55807eb1fa633628a20d6e525027f2a0/src/Scriban/Syntax/Statements/ScriptForStatement.cs#L104-L156)

doc/language.md
```diff
@@ -1151,7 +1151,7 @@ Allows to start the iteration of the loop at the specified zero-based index:
 
 > **input**
 ```scriban-html
-{{~ for $i in 4..9 offset:2 ~}}
+{{~ for $i in 4..9 | array.offset:2 ~}}
```
xoofx, line 1154:
```suggestion
{{~ for $i in (4..9) offset:2 ~}}
```

doc/language.md
```diff
@@ -1169,7 +1169,7 @@ Limits the iteration of the loop to a specified count:
 
 > **input**
 ```scriban-html
-{{~ for $i in 4..9 limit:2 ~}}
+{{~ for $i in 4..9 | array.limit:2 ~}}
```
xoofx, line 1172:
```suggestion
{{~ for $i in (4..9) limit:2 ~}}
```

doc/language.md
```diff
@@ -1185,7 +1185,7 @@ Reverses the iteration of the elements:
 
 > **input**
 ```scriban-html
-{{~ for $i in 1..3 reversed ~}}
+{{~ for $i in 1..3 | array.reversed ~}}
```
xoofx, line 1188:
```suggestion
{{~ for $i in (1..3) reversed ~}}
```

PR #491, opened 2023-02-27
restore net6.0

tibel:
.NET 6.0 is still supported LTS version.
Please restore .NET 6.0 that was removed in 6c6d419d8cb89ed3a9b86f85da028faefeadc643

### DIFF
--- a/src/Scriban/Scriban.props
+++ b/src/Scriban/Scriban.props
@@ -7,7 +7,7 @@
     <LangVersion>8.0</LangVersion>
     <DefineConstants>$(DefineConstants);SCRIBAN_PUBLIC</DefineConstants>
     <Authors>Alexandre Mutel</Authors>
-    <TargetFrameworks>netstandard2.0;net7.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net6.0;net7.0</TargetFrameworks>
     <PackageTags>templating;handlebars;liquid</PackageTags>
     <PackageIcon>scriban.png</PackageIcon>
     <PackageReadmeFile>readme.md</PackageReadmeFile>


xoofx:
Just to understand, what is the problem with `.NET Standard 2.0` the case of .NET 6.0?

tibel:
When targeting .NET 6.0 you I get those old package references now.
``` xml
  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETStandard'">
    <PackageReference Include="Microsoft.CSharp" Version="4.5.0" />
    <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.0" />
  </ItemGroup>
```

This was not the case with Scriban 5.6.0 (see https://www.nuget.org/packages/Scriban/5.6.0#dependencies-body-tab).

xoofx:
> When targeting .NET 6.0 you I get those old package references now.

But these are part of .NET 6.0, so when you install Scriban in a .NET 6.0 project, you won't see these packages.

Not sure why even .NET 5/.NET 6 was added, maybe I thought that was a problem about having these dependencies, but actually, they should not be a problem 🤔

tibel:
> Not sure why even .NET 5/.NET 6 was added, maybe I thought that was a problem about having these dependencies, but actually, they should not be a problem 🤔

I had no issue yet. Only thought that removing a supported .NET version (target framework) was an accident.

Please feel free to close this pull request.

xoofx:
> I had no issue yet. Only thought that removing a supported .NET version (target framework) was an accident.
> Please feel free to close this pull request.

Closing for now. I will probably remove net7.0 as it might confuse as well.
If you find a problem, I can revisit this, it's not a big deal.
Thanks!

Issue #508, opened 2023-06-09
case-insensitive comparison of string?


joshidp:
How can we do case-insensitive of string? I am trying below snippet but its doing case sensitive comparison.

`{{valueA!=valueB}}`

Thanks

ranger-turtle:
I think it's possible and reasonable to make the built-in function making case-insensitive comparison.

For now, you can use this workaround:
`"Left".downcase == "LEFT".downcase`
or
`"Left".upcase == "LEFT".upcase`

ranger-turtle:
> How can we do case-insensitive of string?

You do not need to use workaround I have written in this thread. Builtin `string.equals_ignore_case` is added to the newest version.

PR #509, opened 2023-06-10
Add EqualsIgnoreCase string builtin function

ranger-turtle:
I have added EqualsIgnoreCase string builtin function which would be faster way for case insensitive string comparison than the workaround I have mentioned in my comment in #508 .


binarycow:
(Note - I'm not the maintainer of Scriban - just another occasional contributer) 

You beat me to it!

While you're at it, I have a suggestion, if you have the time...

Add an overload for the normal equals method that accepts `StringComparison`.

Could also do that for all of the other methods, where the .NET method also accepts `StringComparison`. 

IIRC, Scriban supports optional parameters, so it would be as easy as copy/pasting the `StringComparison` parameter in three places for each method:

- Parameter list
- Method call
- XML docs  

If Scriban doesn't support optional parameters, then.... Maybe I'll add that?

ranger-turtle:
I considered making general `equals` builtin function with `StringComparison` type parameter but I decided to make the `EqualsIgnoreCase` function instead, since case sensitive comparison can be accomplished using `==` operator and I found making `equals` unnecessary.

However, I think I may add an optional parameter to such functions, like `StartsWith` or `Contains`. I will wait for @xoofx 's answer to ensure if this PR makes sense according to him.

src/Scriban/Functions/StringFunctions.cs
```diff
@@ -305,6 +305,25 @@ public static bool EndsWith(string text, string value)
...
 
+        /// <summary>
+        /// Returns a boolean indicating whether the input string is equal to specified string 'value'. Comparison is case insensitive.
+        /// </summary>
+        /// <param name="text">The input string</param>
+        /// <param name="value">The string to compare</param>
+        /// <returns><c>true</c> if `text` is equal to string `value`, ignoring case</returns>
+        /// <remarks>
+        /// ```scriban-html
+        /// {{ "Scriban" | string.ends_with "SCRIBAN" }}
```
xoofx, line 316:
```suggestion
        /// {{ "Scriban" | string.equals_ignore_case "SCRIBAN" }}
```

xoofx:
Adding `string.equals_ignore_case` is fine. Tests are failing (but due to the flood of logs, can't see it easily in GitHub, you need to check this locally)

I would prefer not to change existing methods. For 6+ years nobody has been looking for configurable string comparisons, so this is not necessary to change all these. Users can always create custom methods on their side.

PR #518, opened 2023-09-02
Add string interpolation

### LINKED ISSUE #476: Feature request: string interpolation
Most modern languages have string interpolation i.e. possibility to put the expressions into string **without** use of concatenation.
Interpolated strings look like this in C#: `$"Number: {number + 3}"`
Would it be possible to add this kind of syntactic sugar?

ranger-turtle:
This is my first attempt to add string interpolation to Scriban.
I could not make the Parser and Lexer to discard `$` character fully, since it is needed by `ScriptPrinter`. Maybe I have overlooked something but I am not sure.
It might close #476 .


src/Scriban.Tests/TestFiles/020-interpolation/020-interpolation.txt
```diff
@@ -0,0 +1,7 @@
+{{ $"Scriban {4 + 1}.8 can interpolate strings now" }}
```
xoofx, line 1: Add more tests with:
- Multiple string interpolation in from a same string (e.g `$"Hello {a} {b} {c}"`) and other variations (interpolation starting from the beginning e.g `$"{a}"`)
- Add escape handling for `{` (e.g `$"This is not \{an interpolation}"`)

src/Scriban/Parsing/Lexer.cs
```diff
@@ -1037,7 +1068,14 @@ private bool ReadCode()
                     }
 
                     bool specialIdentifier = c == '$';
-                    if (IsFirstIdentifierLetter(c) || specialIdentifier)
+                    char nextChar = PeekChar(1); //Used for checking if the dollar is before the string
+                    if (specialIdentifier &&  (nextChar == '"' || nextChar == '\''))
+                    {
+                        NextChar(); //Skip '$'
```
xoofx, line 1074: We probably don't want to skip `$` here because we want to capture the `start` position correctly from `ReadInterpolatedString` below

src/Scriban/Parsing/Lexer.cs
```diff
@@ -1620,6 +1658,149 @@ private void ReadVerbatimString()
...
+                        case '\'':
+                        case '"':
+                        case '\\':
+                        case '{':
+                        case 'b':
+                        case 'f':
+                        case 'n':
+                        case 'r':
+                        case 't':
+                        case 'v':
```
xoofx, line 1704: Add escape handling for `\{`

src/Scriban/Parsing/Lexer.cs
```diff
@@ -1620,6 +1658,149 @@ private void ReadVerbatimString()
...
+                    AddError($"Unexpected end of file while parsing a string not terminated by a {startChar}", end, end);
+                    return;
+                }
+                else if (c == '{') //If interpolation beginning is read
+                {
+                    end = _position;
+                    // if interpolation is starting
+                    if (!readingInterpolation)
+                    {
+                        start = new TextPosition(start.Offset - 1, start.Line, start.Column - 1); //For '$' inclusion
```
xoofx, line 1767: Remove - 1 once `NextChar` is done after `start`

src/Scriban/Parsing/Parser.Terminals.cs
```diff
@@ -311,6 +316,10 @@ private ScriptLiteral ParseString()
                     ? ScriptLiteralStringQuoteType.SimpleQuote
                     : ScriptLiteralStringQuoteType.DoubleQuote;
 
+            // I had had to do so because of the printer who needs to know that even pure strings may have leading '$' sign
```
xoofx, line 319: That probably can be removed once $ is included as part of the start/end.

src/Scriban/Parsing/Parser.Terminals.cs
```diff
@@ -396,6 +405,125 @@ private ScriptLiteral ParseString()
...
+
+            var end = Current.End.Offset;
+            for (int i = begin; i < end; i++)
+            {
+                var c = text[i];
+                // Handle escape characters
+                if (text[i] == '\\')
+                {
+                    i++;
+                    switch (text[i])
```
xoofx, line 446: Add escape handling `\{`

src/Scriban/Parsing/Lexer.cs
ranger-turtle: I have actually added but I have forgotten to add several test before last commit and after amend, I have got a merge conflict. Being not experienced with Git, it apparently did not upload these changes after rebase.

ranger-turtle: In other hand, I needed to handle it in regular strings, because interpolated strings without any interpolated expression are treated as regular strings.

ranger-turtle:
I am testing the interpolation begin escape and I have found that when I use the escape without adding any interpolated expression, there is an error, since the interpolated string is treated as the regular string and it does not accept the escaped `{`. With leading `$`, it would look weird that user could not escape `{` in the supposedly interpolated string. What is more important, user knows nothing what tokens are used under the hood. I know two solutions:

- Making the interpolated string to be actually interpolated again when at least one `{` is found. However, I need to create additional token;
- Implementing escape for regular strings too. I have applied it to the both types of string and test `"400-builtins\\470-html.txt"` failed for the first time after such change. I came to the conclusion that the implementation escapes will break the backward compatibility with the previous versions.

ranger-turtle:
Besides, I still considered Ruby-style interpolation in meantime and now I think this would be better idea, since:

- Syntax is very Ruby-like and people experienced with Ruby would be confused with the difference of the interpolation syntax in Ruby and the Scriban. If I speak, both C# and Ruby syntaxes are readable for me;
- I would not need to change interpolated string to regular one and to escape `{` and `}`, since interpolation expression begins from the combination of two characters instead of one.

I respect your opinion about interpolation syntax and your time but we develop Scriban for people, not for ourselves. And the change will remain mostly my business after all.

ranger-turtle:
Besides, I still considered Ruby-style interpolation in meantime and now I think this might be better idea, since:

- I would not need to change interpolated string to regular one and to escape { and }, since interpolation expression begins from the combination of two characters instead of one.
- Syntax is fairly Ruby-like and people experienced with Ruby would be confused with the difference of the interpolation syntax in Ruby and the Scriban. In my opinion, both C# and Ruby interpolation syntaxes are readable and natural. Only difference will be that both of the quote types will be supported, but it is much more subtle;

I respect your opinion about interpolation syntax and your time but we develop Scriban for people who use it, not for only both of us. And the change will remain mostly my business after all.

ranger-turtle:
Besides, I still considered Ruby-style interpolation in meantime and now I think this might be better idea, since:

- I would not need to change interpolated string to regular one and to escape `{` and `}`, since interpolation expression begins from the combination of two characters instead of one.
- Syntax is fairly Ruby-like and people experienced with Ruby would be confused with the difference of the interpolation syntax in Ruby and the Scriban. In my opinion, both C# and Ruby interpolation syntaxes are readable and natural. Only difference will be that both of the quote types will be supported, but it is much more subtle;

I respect your opinion about interpolation syntax and your time but we develop Scriban for people who use it, not for only both of us. And the change will remain mostly my business after all.

xoofx:
> I have found that when I use the escape without adding any interpolated expression, there is an error, since the interpolated string is treated as the regular string

Don't treat it as a regular string, but as an interpolated string that doesn't have interpolations. That will avoid entirely the problem. Roundtrip of Scriban text should be preserved anyway, so you would have to keep this information around.

>  I have found that when I use the escape without adding any interpolated expression, there is an error, since the interpolated string is treated as the regular string

No need to change your PR on that aspect. Your PR is close to completion really.

>  I respect your opinion about interpolation syntax and your time but we develop Scriban for people who use it, not for only both of us.

Just to clarify, all the OSS I develop, that's primarily for my personal usage 🙂. I don't develop things for others, but I'm willing to share with others if that is useful and it is not a huge burden for me to maintain. If you contribute to this repo, it means that it is something I'm ok to "support", you can go away and do something else in your life, I will still support it.

lofcz:
Please include tests with nested interpolation `{{ $"Scriban { true ? $"yes {4 + 1}" : "no" } can interpolate strings now" }}`

xoofx:
@ranger-turtle Thanks for the changes, much appreciated, it's looking good. Could you push the "Resolve Conversation" buttons above for the items that you have already resolved so that I can reason about what is left.

ranger-turtle:
> @ranger-turtle Thanks for the changes, much appreciated, it's looking good. Could you push the "Resolve Conversation" buttons above for the items that you have already resolved so that I can reason about what is left.

Done. I think you need to check if the error messages I have added in last two commits are friendly enough.

ranger-turtle:
@xoofx You're welcome. Do not forget about `EqualsIgnoreCase` builtin (see #509 for reminder).

xoofx:
> Do not forget about `EqualsIgnoreCase` builtin (see #509 for reminder).

It was already published for [5.8.0](https://github.com/scriban/scriban/releases/tag/5.8.0) 🙂

ranger-turtle:
> > Do not forget about `EqualsIgnoreCase` builtin (see #509 for reminder).
> 
> It was already published for [5.8.0](https://github.com/scriban/scriban/releases/tag/5.8.0) 🙂

Yes, but there is no info about that builtin in documentation.

xoofx:
> Merci, but there is no info about that builtin in documentation.

Oh, ok different topic. There is an exe in the solution to generate such doc, it's not done manually. No problem I will verify the doc.

ranger-turtle:
> Oh, ok different topic. There is an exe in the solution to generate such doc, it's not done manually. No problem I will verify the doc.

That is why I did not add the info myself. 🙂

xoofx:
Btw @ranger-turtle, I have updated Scriban in [kalk](https://github.com/xoofx/kalk) and String interpolation works out of the box 🙂 

image

ranger-turtle:
@xoofx Just like other Scriban-based software, including MakiSei, one of the my projects. Merci.

NeVeSpl:
Great job! AST also works out of the box, and all unit tests that I have in NTypewriter passed without any problem. 
image

PR #528, opened 2024-01-02
Add nuget .editorconfig

viceroypenguin:
### Problem
When using `<PackageScribanIncludeSource>true</PackageScribanIncludeSource>`, the imported .cs files are analyzed according to the analyzer rules of the local project. This is a problem when consumers have different style settings, or even custom analyzers, for which Scriban code is not valid.

### Solution
This PR adds a `.editorconfig` to the Nuget package and links it via the Scriban.targets. When done this way, the Scriban files are still included for compiling, but are excluded from analysis.


xoofx:
Thanks!

PR #533, opened 2024-02-12
Implements Full Support for System.Text.Json - Closes #527

### LINKED ISSUE #527: Add Support for System.Text.Json.JsonElement
## Current behavior

Wenn using `System.Text.Json` deserializing to `Dictionary<string, object>` does not work, because it produces 
`Dictionary<string, JsonElement>` and `JsonElement` can not be added directly to `ScriptObject` because JsonElement
has no public properties to be exposed in the scriban runtine. 

**See the following Example to understand how scriban handles JsonElement:**

```csharp
var model = System.Text.Json.JsonSerializer.Deserialize<JsonElement>("""{ model: { "foo": "bar" } }""");
var template = Template.Parse("{{ model | object.keys }}");
var result = template.Render(model);
// actual result -> ["value_kind"]
// expected result -> ["foo"]
```

## Expected behavior

It should be possible to add `JsonElement` to a `ScriptObject` or `ScriptArray` while exposing the actual JSON fields to the runtime.

## Motivation for Change

I think providing first class support for .NET build-in JSON types makes scriban a even more powerful engine.

## How I solved it

To solve the problem I wrote a extension method for ScriptObject where I loop through the JsonElement and add all elements to a ScriptObject or ScriptArray.

**Working Example:**

```csharp
var model = System.Text.JsonSerializer.Deserialize<JsonElement>("""{ model: { "foo": "bar" } }""");
var template = Template.Parse("{{ model | object.keys }}");
var scriptObject = new ScriptObject();
scriptObject.Import(model);
var result = template.Render(scriptObject);
// result -> ["foo"]

public static class Extensions {
  public static void Import(this IScriptObject script, JsonElement json)
  {
      if (json.ValueKind is not JsonValueKind.Object) throw new ArgumentOutOfRangeException($"Unsupported object type `{json.ValueKind}`. Expecting Json Object.");
  
      AddObject(json, script);
  
      object? Value(JsonElement model)
      {
          return model.ValueKind switch {
              JsonValueKind.Array => AddArray(model, new ScriptArray()),
              JsonValueKind.Object => AddObject(model, new ScriptObject()),
              JsonValueKind.False => false,
              JsonValueKind.True => true,
              JsonValueKind.Null => null,
              JsonValueKind.Number => model.GetDecimal(),
              JsonValueKind.String => model.GetString(),
              JsonValueKind.Undefined => null,
              _ => throw new ArgumentOutOfRangeException()
          };
      }
  
      IScriptObject AddObject(JsonElement model, IScriptObject obj)
      {
          foreach (var property in model.EnumerateObject()) {
              obj.SetValue(property.Name, Value(property.Value), false);
          }
  
          return obj;
      }
  
      ScriptArray AddArray(JsonElement model, ScriptArray array)
      {
          foreach (var value in model.EnumerateArray()) {
              array.Add(Value(value));
          }
  
          return array;
      }
  }
}
```

## API Proposal

```csharp
class ScriptObjectExtensions {
  public static void Import(this IScriptObject script, JsonElement json, MemberFilterDelegate filter = null, MemberRenamerDelegate renamer = null) {
    if (json.ValueKind is not JsonValueKind.Object) throw new ArgumentOutOfRangeException($"Unsupported object type `{json.ValueKind}`. Expecting Json Object.");
    ...
  }
  public static void Import(this IScriptObject script, object obj, MemberFilterDelegate filter = null, MemberRenamerDelegate renamer = null) {
    if (obj is JsonElement json) {
      script.Import(json, filter, renamer);
    }
    ...
  }
}
class ScriptObject {
  public void Add(string key, JsonElement value);
}
class ScriptArray {
  public void Add(JsonElement value);
}
```

## Pull-Request?

I would provide a pull request.

r-Larch:
This update introduces comprehensive support for System.Text.Json within the Scriban runtime. All modifications are confined to .NET Core build targets, delineated by the #if NET compiler directive. Documentation updates are included in this pull request.

✅ All tests are green.

## New Public API

- `Template.Render(JsonElement model)` - New overload to render a template directly using a JsonElement as model.
- `ScriptObjectExtensions.Import(this IScriptObject script, JsonElement json)` - New overload to import from JsonElement.
- `ScriptObject.From(JsonElement model)` - New static overload to create a script object from JsonElement.
- `ScriptObject.Add(string key, JsonElement value)` - New overload to add a JsonElement value
- `ScriptObject.SetValue(string member, JsonElement value, bool readOnly)` - New overload to member value to JsonElement
- `ScriptObject.TrySetValue(TemplateContext context, SourceSpan span, string member, JsonElement value, bool readOnly)`
- `ScriptArray<T>.TrySetValue(TemplateContext context, SourceSpan span, string member, JsonElement value, bool readOnly)`
- `ScriptArray.Add(JsonElement item)` - New overload to add a JsonElement value.

## New Internal API

- `JsonElementExtensions.ToScriban(this JsonElement model)` - Helper to convert a JsonElement to a scriban value.
- `ScriptObjectExtensions.AddJsonObject(this IScriptObject obj, JsonElement json)` - Helper to add JsonElement object to ScriptObject.
- `ScriptObjectExtensions.AddJsonArray(this ScriptArray array, JsonElement json)` - Helper to add JsonElement array to ScriptArray.

## New Runtime Functions

- `object.from_json <string>` - Converts a JSON to scriban value.
- `object.to_json <any>` - Converts the scriban value to its JSON representation.

## New Unit Tests

- `TestScriptObjectJson` - Tests all modified ScriptObject overloads.
- `TestScriptArrayJson` - Tests all modified ScriptArray overloads.
- `TestModelWithJsonElement` - Tests various combinations of typed models with JsonElement, boxed and unboxed.
- `TestTemplateRenderJson` - Tests the new Template.Render overload.
- `TestObjectFunctions` - Tests the new runtime JSON functions.

## Notes

The implementation and behavior of this feature have met our expectations, with the exception of the `ScriptArray.Add` method's new behavior, which introduces a minor inconsistency. Consider the following example for clarification:

```csharp
var array = new ScriptArray();
array.Add(jsonElement)  // works fine by converting jsonElement to a scriban value.
array.IndexOf(jsonElement) // Fails to find the element, returning -1, since the value is no longer a JsonElement and thus cannot be matched.
```

I'd appreciate your thoughts on the ScriptArray.Add behavior—should I remove it? Although it's not strictly necessary, it was included as an enhancement.


src/Scriban/Runtime/ScriptObject.cs
```diff
@@ -86,6 +90,9 @@ public ScriptObject(int capacity, bool? autoImportStaticsFromThisType, IEquality
         void IDictionary.Add(object key, object value)
         {
             this.AssertNotReadOnly();
+#if NET
+            value = value is JsonElement json ? json.ToScriban() : value;
```
xoofx, line 94: Do we really need to support dynamic support? Wouldn't the requirement of using `ScriptObject.Import` be enough? It feels quite specific and it would open the door to `is Yaml`, `is Xml`...etc.

r-Larch: In my latest commit, I removed all unnecessary JSON handling overloads and conditional statements. I retained only the elements necessary for the `TestModelWithJsonElement.cs` test cases, as I need this feature for my project where I utilize Scriban. I would really like to maintain this functionality. What are your thoughts?

xoofx: > I retained only the elements necessary for the TestModelWithJsonElement.cs test cases, as I need this feature for my project where I utilize Scriban.

I prefer that Scriban supports import/export of JSON to/from `ScriptObject`, but I really would like to avoid introducing a special case in the accessor of `ScriptObject`. The conversion of a `JsonElement` should only happen at import time, not implicitly at runtime on your behalf, otherwise it is making a significant exception. If we were starting to do that we would have to change also the `TypedObjectAccessor` and other accessors.

I don't think that it will change much `TestModelWithJsonElement.cs` as I believe import time should cover that. And if it wouldn't for your use case, you can always inherit from `ScriptObject` in your app and override set/get value if you really want to have such behavior for your app.

r-Larch: So, if I understand you correct then removing every addition to ScriptObject except the `From` static method and implementing the JsonElement handling in the `Import` extension would be ok, right?

xoofx: Yep and the tests should still work

src/Scriban/Runtime/ScriptObject.cs
```diff
@@ -285,12 +289,18 @@ public virtual bool TrySetValue(TemplateContext context, SourceSpan span, string
         {
             if (!CanWrite(member)) return false;
             this.AssertNotReadOnly();
+#if NET
```
xoofx, line 292: Could you remove these lines? as they should not be required for Import

r-Larch:
In my latest three commits, I made the following changes:
- I moved the JSON handling functionality to the Import extensions, as we discussed.
- I also removed the `ScriptObject.From(JsonElement)` and `Template.Render(JsonElement)` explicit methods because, as you mentioned, it opens the door for `Yaml`, `XML`, and so on. Initially, these methods were included as an optimization to avoid requiring the runtime to box the `JsonElement` ValueType. However, I now believe this optimization is not critically important. 

Now the changes are restricted to only what is needed to support JsonElement importing.
What do you think?

xoofx:
Thanks for the work, I appreciate!

xoofx:
It will be part of a coming release. I'm waiting for other pending PRs to make it in.

Deltids:
When can we expect 5.10 / 6.0? :)

PR #535, opened 2024-02-16
Add array.aggregate function

### LINKED ISSUE #532: Array reduction
Hi,

I'm looking for a way to reduce arrays, e.g. I was hoping for something like this:

```
total_length = ["Apple", "Banana", "Computer", "Mobile Phone", "Orange", "Sofa", "Table"] | array.each @string.size | array.reduce @math.plus
```

Or in the case of summing, just 
```
total_length = ["Apple", "Banana", "Computer", "Mobile Phone", "Orange", "Sofa", "Table"] | array.each @string.size | math.plus
```

But I can't find an obvious way. Is array reduction missing?

igitur:
Fixes #532 

Instead of a reduce function, I implemented `array.aggregate`, that also accept an initial seed value. I could add more tests if you want to. Just point me to the applicable spot in the test suite where this would be required.

### DIFF
--- a/doc/builtins.md
+++ b/doc/builtins.md
@@ -20,6 +20,7 @@ Array functions available through the object 'array' in scriban.
 
 - [`array.add`](#arrayadd)
 - [`array.add_range`](#arrayadd_range)
+- [`array.aggregate`](#arrayaggregate)
 - [`array.compact`](#arraycompact)
 - [`array.concat`](#arrayconcat)
 - [`array.cycle`](#arraycycle)
@@ -99,6 +100,38 @@ The concatenation of the two input lists
 > **output**
 ```html
 [1, 2, 3, 4, 5]
+```
+
+[:top:](#builtins)
+### `array.aggregate`
+
+```
+array.aggregate <list> <seed> <function>
+```
+
+#### Description
+
+Applies an accumulator function on the list.
+
+#### Arguments
+
+- `list`: An input list
+- `seed`: The initial accumulator value
+- `function`: The function to apply to each item in the list
+
+#### Returns
+
+The transformed final accumulator value.
+
+#### Examples
+
+> **input**
+```scriban-html
+{{ [4, 5, 6, 7] | array.aggregate 5 @math.plus}}
+```
+> **output**
+```html
+27
 ```
 
 [:top:](#builtins)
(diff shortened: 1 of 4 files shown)


igitur:
And if you're happy with this approach, I'd like to implement more LINQ-ish array functions. `array.sum`, `array.average`, `array.all`, `array.any`, `array.min`, `array.max` and maybe `array.groupby`.

xoofx:
Oh definitely, you can add them as part of this PR

r-Larch:
Hi,

I really appreciate these additions; great work!

I reviewed the `array.aggregate` implementation and believe it could be improved by making the seed value optional. Making the seed value optional can enhance flexibility and usability for cases where the initial value isn't necessary or when the user prefers the aggregation to start directly with the elements of the array. Specifically, if no seed value is provided, it would be beneficial to automatically use the first item of the list as the `current` value and the second item as the `next` value. This approach simplifies the function's usage for common aggregation tasks, such as summing values or concatenating strings, where the natural starting point is the first element of the array. It also aligns with the behavior of similar functions in other programming languages, making it more intuitive for users transitioning from different backgrounds.

What do you think?

igitur:
@r-Larch I played around with that idea. The order if the 2 parameters would need to swap from seed & function to function and optional seed. Right? But when I tried that with the seed provided, the function consumes the seed instead of it being passed into `array.aggregate`. Not sure if it's a bug or by design.

igitur:
Maybe just easier to implement `array.reduce`, which wouldn't have the seed parameter.

r-Larch:
> @r-Larch I played around with that idea. The order if the 2 parameters would need to swap from seed & function to function and optional seed. Right? But when I tried that with the seed provided, the function consumes the seed instead of it being passed into `array.aggregate`. Not sure if it's a bug or by design.

Hmm.. for me it looks like a bug but I don't know.

Does it work with parameters like this: (seed optional before func)
|Method|Desc|
|-|-| 
| `array.aggregate <func(c, n)>` | no seed, just a func param | 
| `array.aggregate <seed> <func(c, n)>` | seed + func | 

In C#
```csharp
public static object Aggregate(TemplateContext context, SourceSpan span, IEnumerable list, object? arg1, object? arg2);
if (arg1 is IScriptCustomFunction) ...
if (arg2 is IScriptCustomFunction) ...
```

xoofx:
Closing for now, as it got stale, feel free to reopen once things are better settled. Thanks!

PR #545, opened 2024-04-25
Update runtime.md: add documentation on getting variables from global

BenJBailey:
Clarified how to get variables anywhere and at the top of the global stack from within a function.


xoofx:
Thanks!

PR #547, opened 2024-05-31
feat: Support targetframework .net 8

### LINKED ISSUE #546: Support library for target framework .net 8
.net 7 is Out of support versions.

https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core

image

eoehen:
close #546

### DIFF
--- a/src/Scriban/Scriban.csproj
+++ b/src/Scriban/Scriban.csproj
@@ -2,5 +2,6 @@
   <Import Project="Scriban.props" />
   <PropertyGroup>
     <PackageId>Scriban</PackageId>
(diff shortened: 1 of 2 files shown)


eoehen:
I wasn't sure if I was allowed to remove .net 7.

xoofx:
Hey, strictly speaking it's not super important because supporting net7.0 defaults to support net8.0.

I would like to check more broadly what needs to be updated to net8.0 in the project (there are several executables...etc.)

pvanroos:
I tried a simple console project targeting .NET 9 and it does not work.

pvanroos:
I'll try again.  I swapped it out for Fluid in the interim.

PR #554, opened 2024-06-26
Fix a bug in bug in DictionaryAccessor

SergeyXa:


xoofx:
Yep, good catch, many of these methods don't have proper tests, and you are the first in several years, so they were not that used it seems 😅 

Could you add a small one for this case in e.g [TestRuntime.cs](https://github.com/scriban/scriban/blob/master/src/Scriban.Tests/TestRuntime.cs)?

PR #557, opened 2024-07-01
Fix: If statement comparing string and null value results in object reference error

BenJBailey:
When comparing a null variable with a string, an error is thrown.  

For example:
```liquid
{% assign nothing = null %}
{% assign bar = 'foo' %}
{% if nothing > bar %}
  Not equal
{% endif %}
```

Produces `Object reference not set to an instance of an object.`

The conditional logic is checking for value of type `strings before it checks if a value is `null`.

It would make sense to check if values are `null` and run `CalculateOther(...)` if either left or right value is `null`.

If `null` and `null` can be compared without an error, `"a"` and `null` should also be able to be compared without an error.

This pull request adds a null check before determining the type of either value when comparing values in an if statement.


xoofx:
Thanks!

PR #561, opened 2024-08-07
feat: fix #555 Error object.to_json with date or timespan

### LINKED ISSUE #555: Error `object.to_json` with `date` or `timespan`
The following code produces an error:

```f#
{{ "1024-06-26" | date.parse | object.to_json }}
// -> error : CurrentDepth (1000) is equal to or larger than the maximum allowed depth of 1000. 
// -> Cannot write the next JSON object or array.
// Expected Result: "1024-06-26T00:00:00+02:00"
```

The error comes from `DateTime` being interpreted as 'object' and not as a value.

## A possible fix 

This and probably all such cases could be fixed with the following change:
- Ensure that all objects implementing `IFormattable` are being threaded as a value rather than an object.

https://github.com/scriban/scriban/blob/38df14946721238d2f4badc7c0b6e8a1c594e23c/src/Scriban/Functions/ObjectFunctions.cs#L481-L532

```diff
var type = value?.GetType() ?? typeof(object);
if (
    value is null ||
    value is string ||
    value is bool ||
-    type.IsPrimitiveOrDecimal()
+    type.IsPrimitiveOrDecimal() ||
+    value is IFormattable
)
{
    JsonSerializer.Serialize(writer, value, type);
}
else if (value is IList || type.IsArray) {
```

r-Larch:
Add serialization support for all types deriving from IFormattable. All System types including time values like DateTime, TimeSpan, DateOnly and other special values like Uri, Guid, BigInteger, etc.. derive from IFormattable and are therefore compatible.

This PR improves, overwrites, and replaces PR-https://github.com/scriban/scriban/pull/558 
- The previous solution only handles/adds support for DateTime and fails to handle eg. TimeSpan, Guid, ...
- The current PR adds support for a wide range of types. [See All IFormattable Derived Types](https://learn.microsoft.com/en-us/dotnet/api/system.iformattable?view=net-8.0)

Breaking Changes

Both, the previous and the current PR use ISO8601 format for `DateTime` encoding.
So everything should work fine.
But the new PR slightly changes the actual ISO8601 representation.
eg. 

```c#
[Test]
public void JsonSerializer_produces_iso8601_with_trimmed_zeros()
{
    var date = new DateTime(2024, 08, 07, 15, 27, 12, 234, DateTimeKind.Utc);

    // old behavior:
    var iso8601Result = date.ToString("O");
    Assert.AreEqual("2024-08-07T15:27:12.2340000Z", iso8601Result);

    // new behavior: useless zeros are removed
    var jsonResult = JsonSerializer.Serialize(date);
    Assert.AreEqual("\"2024-08-07T15:27:12.234Z\"", jsonResult);
}
```

### DIFF
--- a/src/Scriban.Tests/TestJsonSupport/TestTypeSerialization.cs
+++ b/src/Scriban.Tests/TestJsonSupport/TestTypeSerialization.cs
@@ -1,0 +1,55 @@
+// Copyright (c) Alexandre Mutel. All rights reserved.
+// Licensed under the BSD-Clause 2 license.
+// See license.txt file in the project root for full license information.
+
+using Newtonsoft.Json.Linq;
+using System;
+using System.Globalization;
+using System.Text.Json;
+
+namespace Scriban.Tests.TestJsonSupport;
+
+[TestFixture]
+public class TestTypeSerialization {
+    public static object[][] TestValues = [
+        [(DateTime) new DateTime(2024, 08, 07, 15, 27, 0), "\"2024-08-07T15:27:00\""],
+        [(TimeSpan) TimeSpan.FromTicks(123434563), "\"00:00:12.3434563\""],
+        [(DateOnly) DateOnly.FromDateTime(new DateTime(2024, 08, 07)), "\"2024-08-07\""],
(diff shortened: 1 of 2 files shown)


xoofx:
Thanks, I'm on holidays, will come back only in ~2 weeks

Issue #165, opened 2019-04-24
array.map but for functions, not properties


igitur:
Hi,

I'm looking for a way to map a function on an array. I see that the current [array.map](https://github.com/lunet-io/scriban/blob/master/doc/builtins.md#arraymap) function can map only to a property of the object. I would like to apply an arbitrary function on an array without having to iterate through the array manually with e.g. a `for` loop. I'm kind of looking for the equivalent of C#'s `array.Select(o => SomeFunction(o)).ToArray()`.

Is this possible?

thanks

xoofx:
Unfortunately, there is nothing like this. It should be possible to bring `array.select` and pass a function to it. PR welcome.

igitur:
Ok, I'll attempt a PR.

Do you agree with this syntax? I believe the function will have to be passed using the `@` operator, right?

```elm
{{ 
func length
   ret string.size $0
end

fruit = ["orange", "banana", "apple", "pear", "strawberry"]
fruit | array.select @length }}
```

xoofx:
yep looks good 👍 

I don't remember if there was already a case where we expose a native function which can take a function pointer, might be a bit more involving to get it working

igitur:
Just as an aside, my actual use case is to have an array of dates and in some way get them joined and formatted to end up with a single string (I'm using scriban to build up complex SQL queries):
```sql
where effective_date in ('2019-01-01','2019-01-02','2019-01-03')
```

I could iterate through the dates and format them one-by-one, but then I have to deal with not having a trailing comma. Or I could push the formatted dates to a new array and then use `array.join`, but that's too verbose.

xoofx:
> I could iterate through the dates and format them one-by-one, but then I have to deal with not having a trailing comma. Or I could push the formatted dates to a new array and then use array.join, but that's too verbose.

Yeah, apart providing your own function that wraps array.join with a format, there is no builtin solutions for that

igitur:
> might be a bit more involving to get it working

Yes, turns out you're right. This isn't as simple as I initially expected.

MarcusWichelmann:
Are there any plans to implement this feature? I think this would be a great extension to this very cool templating language.

mrpmorris:
I really needed this today, for my C# language templating plugin.

jons-bakerhill:
It looks like this can (now?) be done with Array.Each
`for propertyType in Array.Each(publicProperties, @GetPropertyType) | Array.Sort `

PR #575, opened 2024-11-18
Add array.select function

jason-kingsmill:
Adds a new _array.select_ function, allowing projection of array items into a new form.

Usage included in docs, with added tests.

Solves for #165


jason-kingsmill:
Turns out array.each has the identical functionality

PR #582, opened 2024-12-13
enhancement: string concatenation with null #579

### LINKED ISSUE #579: The string concatenation with null has an error
Please refer to the following code

```scriban
# The value of the name variable comes from the return value of another function and may be null
# For simplicity in testing, it is fixed to null here
name = null

text = "my name is " + name
# Interpolated strings are the same.
# text = $"my name is {name}"

# Output the text variable, its value is: null
print text
```

The same code behaves like this in C#：
```csharp
string name = null
var text = "my name is " + name

# Output the text variable, its value is: my name is 
Console.WriteLine(text)

```

md-frank:
Hello,

I've fixed #579

### DIFF
--- a/src/Scriban.Tests/TestRuntime.cs
+++ b/src/Scriban.Tests/TestRuntime.cs
@@ -1290,6 +1290,23 @@ Tax: {{ 7 | match_tax }}";
             TextAssert.AreEqual("42-42-42-42", result);
         }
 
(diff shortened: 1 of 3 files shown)


xoofx:
Closing with another fix as per https://github.com/scriban/scriban/issues/579#issuecomment-2550452127

PR #593, opened 2025-03-05
Fix DateTime and Object tests

user-named-void:
A Doc_date test failed because it validates the date.now.year.

### DIFF
--- a/doc/builtins.md
+++ b/doc/builtins.md
@@ -749,7 +749,7 @@ Returns a datetime object of the current time, including the hour, minutes, seco
 ```
 > **output**
 ```html
-2024
+2025
 ```
 
 [:top:](#builtins)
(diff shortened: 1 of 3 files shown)


xoofx:
Not your fault (probably a previous PR), but it seems the JSON version in the code is invalid and has invalid spaces:

https://github.com/scriban/scriban/blob/f8fa7e2a884691a42286912d558a91a57eca87ce/src/Scriban/Functions/ObjectFunctions.cs#L441-L481

If you can correct it as well, that would be great, thank you!

xoofx:
Basically it should be:

```C#
 /// ```html 
 /// {"foo":"bar","baz":[1, 2, 3]} 
 /// true 
 /// null 
 /// ``` 
 /// </remarks> 
 public static string ToJson(TemplateContext context, object value) 
```

user-named-void:
Yes, i fixed the object.to_json test, all test are now succesfull on my system

xoofx:
Thank you for the quick fix!

PR #596, opened 2025-03-05
Fix ScriptRange Add when IList has a fixed size

### LINKED ISSUE #585: .NET9 ScriptRange Add NotSupportedException: Collection was of a fixed size
When migrating to .NET9 we found an issue with the array.add function. Using an array.add on a non existing list we get an error 'Collection was of a fixed size'
`arr = arr | array.add 1`

If the list is null, the array.add function tries to create a new ScriptRange with the new value, but this fails with the NotSupportedException.
https://github.com/scriban/scriban/blob/f8fa7e2a884691a42286912d558a91a57eca87ce/src/Scriban/Functions/ArrayFunctions.cs#L47
Switching back to .NET8 fixes the issue.

user-named-void:
With .NET9 casting the _values to an IList no longer results in a null value. This cast to an fixed size ILIst.
This results in an "Collection was of a fixed size" when trying to add a new item

When the list has a fixed size it will be recreated as an ScriptArray to fixed this, this fixes #585.


user-named-void:
The Test "A400_builtins" already tests the array.add function and it fails without this patch if the test application is run under .net9. I don't know what other tests i should write.

PR #600, opened 2025-03-13
Template code made more customizable

### LINKED ISSUE #599: Advanced custom function questions
Hello,

I am using Scriban to actually parse database queries and I have a few questions on particular points of scriban code (note that i can make PR myself if you agree with proposed changes):

1. I am implementing alternatives to the include function but it seems a method required to handle new lines properly is not accesible from outside scriban. I have looked around the TemplateContext object but can't find a way to replicate the ResetPreviousNewLine() method. Would you be so kind to make it public or to indicate me an alternative to reproduce the behaviour ?

2. I am caching all parsed templates as Template object, is there a reason it would be bad practice ? If not, it would be great if either of the following changes by order of preference could be made:
 - The ITemplateLoader Load method and LoadAsync method could return an object instead of a string. The value of this is that we could return a precached template or string template and adapt the behaviour in IncludeFunction accordingly. Given the usage, I don't think this would introduce much breaking changes but up to you.
 - The template cache dictionary could be provided as a TemplateContext constructor parameter or we could have a setting to provide existing cache instead of filling elements one by one from another dictionary.

3. Finally, the main question of this post: I would like to build a custom function allowing to input an arbitrary number of blocks as parameters. I'm aware of the wrap function but this only allow for a single block to be passed. I m intending to create my own implementation of necessary interfaces of statements and expression onto the function similarly to ScriptFunction but I'm not certain this is possible nor the right approach. Could you advice on this ?
Exemple of the idea:
```
{% block_function [usual params] %}
  block param 1
{% next %}
  block param 2
{% next %}
  block param 3
{% end %}
```
Thanks for your wise input.

julichan:
This fixes #599:
- TemplateContext.ResetPreviousNewLine()
- Moved template processing code from IncludeFunction to TemplateContext and made it customizable.

### DIFF
--- a/src/Scriban/TemplateContext.cs
+++ b/src/Scriban/TemplateContext.cs
@@ -1203,7 +1203,7 @@ namespace Scriban
             return null;
         }
 
-        internal void ResetPreviousNewLine()
+        public void ResetPreviousNewLine()
         {
             _previousTextWasNewLine = false;
         }
--- a/src/Scriban/TemplateContext.cs
+++ b/src/Scriban/TemplateContext.cs
@@ -1207,6 +1207,46 @@ namespace Scriban
         {
             _previousTextWasNewLine = false;
         }
+
+        public Template GetOrProcessTemplate(string templatePath, ScriptNode callerContext)
+        {
+            if (!CachedTemplates.TryGetValue(templatePath, out var template))
+            {
+                template = ProcessTemplate(templatePath, callerContext);
+                CachedTemplates[templatePath] = template;
+            }
+            return template;
+        }
+
+        protected virtual Template ProcessTemplate(string templatePath, ScriptNode callerContext)
+        {
+            string templateText;
+            try
+            {
+                templateText = TemplateLoader.Load(this, callerContext.Span, templatePath);
+            }
+            catch (Exception ex) when (!(ex is ScriptRuntimeException))
+            {
+                throw new ScriptRuntimeException(callerContext.Span, $"Unexpected exception while creating template from path `{templatePath}`", ex);
+            }
+
+            if (templateText == null)
+            {
+                throw new ScriptRuntimeException(callerContext.Span, $"The result of including `{templatePath}` cannot be null");
+            }
+
+            var template = Template.Parse(templateText, templatePath, TemplateLoaderParserOptions, TemplateLoaderLexerOptions);
+
+            // If the template has any errors, throw an exception
+            if (template.HasErrors)
+            {
+                throw new ScriptParserRuntimeException(callerContext.Span, $"Error while parsing template `{templatePath}`", template.Messages);
+            }
+
+            CachedTemplates.Add(templatePath, template);
+
+            return template;
+        }
     }
 
     /// <summary>
--- a/src/Scriban/TemplateContext.cs
+++ b/src/Scriban/TemplateContext.cs
@@ -1208,17 +1208,17 @@ namespace Scriban
             _previousTextWasNewLine = false;
         }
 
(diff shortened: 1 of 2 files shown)


src/Scriban/TemplateContext.cs
```diff
@@ -1203,10 +1203,50 @@ protected virtual IListAccessor GetListAccessorImpl(object target, Type type)
...
+        {
+            if (!CachedTemplates.TryGetValue(templatePath, out var template))
+            {
+                template = ProcessTemplate(templatePath, callerContext);
+                CachedTemplates[templatePath] = template;
+            }
+            return template;
+        }
+
+        protected virtual Template ProcessTemplate(string templatePath, ScriptNode callerContext)
```
julichan, line 1221: I chose the word Process rather than Create because Create did not feel like the right action since it's loading a template from a source and parsing a string into a Template objects. Overrides may not even go that way. I could not find something less general than Process but this feels fitting to me. What do you think ?

src/Scriban/TemplateContext.cs
```diff
@@ -1203,10 +1203,50 @@ protected virtual IListAccessor GetListAccessorImpl(object target, Type type)
             return null;
         }
 
-        internal void ResetPreviousNewLine()
+        public void ResetPreviousNewLine()
         {
             _previousTextWasNewLine = false;
         }
+
+        public Template GetOrProcessTemplate(string templatePath, ScriptNode callerContext)
```
xoofx, line 1211: I think I prefer the name `GetOrCreateTemplate`. `Process` might imply that it is running it or something more than just creating a Template object.

julichan: I made the exact opposite comment with an explanation of my choice on another line. haha.

julichan: To process something refers to a series of actions or steps taken to achieve a specific outcome. This can apply to various contexts, such as converting raw materials into a finished product, handling data, or executing a business operation. In a business context, a process definition is a detailed description of a process, including its start transition and state definitions, expressed through finite guarded recursive specifications.

src/Scriban/TemplateContext.cs
xoofx: Ahah, I said the opposite below. 😅 

I think that Create is easier to reason about. Process almost implies that it is going to render it. This method returns a Template, so it fits well with the generic term "Create".

julichan: May I argue on this, I posted the definition of to process on your post but if you still disagree, I won't oppose to make the change. 🤣

julichan: Sorry, I'm being picky about the choice of word of a function I provide. I can propose the action "Prepare" instead as well if Process really displease you.

src/Scriban/TemplateContext.cs
xoofx: Yeah, naming is hard and while it's debatable, it can be like colors 😁. This function just creates a template. It does not process it. The default implementation parses a text representation to return a Template instance, another implementation could just return an instance from a pool of templates. In the end it creates an instance to put in the cache, and this is usually what people think about in the .NET ecosystem (e.g. [CacheExtensions.GetOrCreate](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.caching.memory.cacheextensions.getorcreate))

julichan: ok you made your point. I'll made an update instead of stalling on this 😄

julichan: done

xoofx:
Thanks!

julichan:
Hell, no! Thanks to you @xoofx !!!!

PR #624, opened 2025-08-18
Add uint calculation

dunds-com:
When having an unsigned integer in the model, it didn't allow calculations.
```cs
var template = Template.Parse("{{ test1 * test2 }}");
var result = template.Render(new
{
	test1 = (uint)55,
	test2 = (uint)80
});

Console.WriteLine(result);
Console.ReadKey(true);
```
> Unsupported types `55/uint` * `80/uint` for binary operation

This PR adds support for uints.

### DIFF
--- a/src/Scriban/Functions/BuiltinFunctions.cs
+++ b/src/Scriban/Functions/BuiltinFunctions.cs
@@ -17,7 +17,7 @@ namespace Scriban.Functions
         /// </summary>
         internal static readonly ScriptObject Default = new DefaultBuiltins();
 
-        public BuiltinFunctions() : base(10)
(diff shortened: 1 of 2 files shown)


dunds-com:
Should I add a scriban file as a unit test?

xoofx:
> Should I add a scriban file as a unit test?

Oh, no, you can add just a test in e.g. [TestRuntime.cs](https://github.com/scriban/scriban/blob/master/src/Scriban.Tests/TestRuntime.cs) similar to what is already done for some other specific cases.

Wondering also if the result should not be `ulong` in the end. BigInteger is quickly quite less practical.

PR #627, opened 2025-09-21
Fix Liquid date output formatting

### LINKED ISSUE #173: Liquid Date Filter returning error
The Liquid date filter is returning an error of 

ScriptRuntimeException: <input>(100,23) : error : Invalid number of parameters `2` for `date` object/function.

Example: 

{{ "March 14, 2016" | date: "%b %d, %y" }} should print Mar 14, 16, however - it is throwing an error in my project. 

{{ "March 14, 2016" }} prints normally as expected. 

Is this an error on my end, or does Scriban have a different method of converting dates in Liquid templates?

meld-cp:
Fixes #173

Added a new function in `DateTimeFunctions` and linked it to the Liquid `date` function.

The Liquid `date` function seems to use the format argument as an output format where as the `DateTimeFunctions.Parse` function uses the pattern as an input format to parse the date.

According to the Liquid docs:
`{{ "March 14, 2016" | date: "%b %d, %y" }}` should result in `Mar 14, 16`

### DIFF
--- a/src/Scriban/Functions/DateTimeFunctions.cs
+++ b/src/Scriban/Functions/DateTimeFunctions.cs
@@ -236,14 +236,61 @@ namespace Scriban.Functions
 
         private static readonly Regex PlusFollowedByNumberRegex = new Regex(@"\+\d");
 
+        private static string ParseCustomFormat(CultureInfo culture, string pattern, out CultureInfo cultureOverride)
+        {
+            cultureOverride = culture;
+            if (pattern == null)
+            {
+                return null;
+            }
+
+            var builder = new StringBuilder();
+            for (int i = 0; i < pattern.Length; i++)
+            {
+                var c = pattern[i];
+                if (c == '%' && (i + 1) < pattern.Length)
+                {
+                    i++;
+                    var format = pattern[i];
+
+                    // Switch to invariant culture
+                    if (format == 'g')
+                    {
+                        cultureOverride = CultureInfo.InvariantCulture;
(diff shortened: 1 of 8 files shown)


src/Scriban/Functions/DateTimeFunctions.cs
```diff
@@ -269,79 +316,50 @@ public static DateTime AddMilliseconds(DateTime date, double millis)
...
-            return result;
+            if (outputPattern is null)
+            {
+                return resultDateTime;
+            }
+
+            return resultDateTime.ToString( outputCustomFormat, outputCulture);
+        }
+
+        public static object ParseWithFormattedOutput(TemplateContext context, string text, string pattern = null, string culture = null)
```
xoofx, line 360: An XML API doc should be added here similar to the `Parse` method.

Then the tool [Scriban.DocGen](https://github.com/scriban/scriban/tree/master/src/Scriban.DocGen) should be ran to update the doc and the tests.

meld-cp: Thanks. 👍 

I was thinking about the new `ParseWithFormattedOutput` function last night, and my understanding is that this new function is now also available to templates via `{{ $.mydate | date.parse_with_formatted_output ... }}` similar to how `date.parse` is.  I find `parse_with_formatted_output` quite a verbose name to be exposed for use in templates so I think I need a better name for it, unless it can be hidden for template use somehow?

PR #635, opened 2025-11-18
Fix #634: ObjectToString has side-effect leading to incorrect escaping with lazy evaluation

### LINKED ISSUE #634: `ObjectToString` has side-effect leading to incorrect escaping with lazy evaluation
Hello,

While using scriban, I can across a problem with `ObjectToString` where the escaping is incorrect.

### Issue reproduction

```C#
var script = """[["a", "b"]] | array.each @(do; ret $0 | array.join("#"); end)""";

var context = new TemplateContext();
var result = Scriban.Template.Evaluate(script, context);
Console.WriteLine(context.ObjectToString(result));
```

### Actual output
```
["\"a\"#\"b\""]
```

### Expected output
```
["a#b"]
```

I think it happens because `context.ObjectToString` is called inside `array.join` when evaluating so both the lazy evaluation and display logic is increasing `_objectToStringLevel`:
https://github.com/scriban/scriban/blob/63bc3d61ae07a05eb1f3e07f352d32f968962a48/src/Scriban/Functions/ArrayFunctions.cs#L395

### Proposed solution

Pass state as parameters to `ObjectToString` (defaults to 0) instead of using instance variables to avoid side-effects

https://github.com/scriban/scriban/blob/63bc3d61ae07a05eb1f3e07f352d32f968962a48/src/Scriban/TemplateContext.Helpers.cs#L80-L81

anya-hichu:
Unfortunately I cannot compile the project due to SDK version mismatch, but here is the proposal I had in mind to fix #634

### DIFF
--- a/src/Scriban/TemplateContext.Helpers.cs
+++ b/src/Scriban/TemplateContext.Helpers.cs
@@ -121,10 +121,10 @@ namespace Scriban
                     var index = LimitToString - _currentToStringLength;
                     if (index <= 0) return string.Empty;
                     str = str.Substring(0, index);
-                    return nested ? $"\"{StringFunctions.Escape(str)}" : (string)value;
+                    return nested ? $"\"{StringFunctions.Escape(str)}" : str;
                 }
 
-                return nested ? $"\"{StringFunctions.Escape(str)}\"" : (string)value;
+                return nested ? $"\"{StringFunctions.Escape(str)}\"" : str;
             }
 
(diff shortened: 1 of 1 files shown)


xoofx:
There is a stack overflow happening in the tests.

> Unfortunately I cannot compile the project due to SDK version mismatch, 

You have to install [.NET 10](https://dotnet.microsoft.com/en-us/download/dotnet/10.0) in order to run the tests locally.

PR #641, opened 2026-03-03
include template - Named array args don't stay as arrays in the template

meld-cp:
This is a fix and tests for the following issue.

---

If we have an include function with 3 args:
```
{{~ include 'named_arguments_promoted_array_types' [1,2,3] x: data y: [4,5,6] ~}}
```

where `named_arguments_promoted_array_types` is:
```
{{ $1 | object.typeof }}::{{ $1 }}
{{ $.x | object.typeof }}::{{ $.x }}=={{ $x | object.typeof }}::{{ $x }}
{{ $.y | object.typeof }}::{{ $.y }}=={{ $y | object.typeof }}::{{ $y }}
```

and `data` is an array of `["one", "two", "three"]`

I think the expected output should be:
```
array::[1, 2, 3]
array::["one", "two", "three"]==array::["one", "two", "three"]
array::[4, 5, 6]==array::[4, 5, 6]
```
(i.e. the type shouldn't change between the `$.x` and `$x` versions)

but instead the named arg types are `object`:
```
array::[1, 2, 3]
array::["one", "two", "three"]==object::["one", "two", "three"]
array::[4, 5, 6]==object::[4, 5, 6]
```

### DIFF
--- a/src/Scriban.Tests/TestJsonSupport/TestScriptArrayJson.cs
+++ b/src/Scriban.Tests/TestJsonSupport/TestScriptArrayJson.cs
@@ -49,6 +49,51 @@ public class TestScriptArrayJson {
     }
 
 
+    [TestCase("""[1, 2, 3]""", "{{ array | object.typeof }}", "array")]
+    [TestCase("""[1, 2, 3]""", "{{ array[0] }}", "1")]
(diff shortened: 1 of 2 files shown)


xoofx:
Thank you for the fix!

PR #643, opened 2026-03-05
Fix `Include` named arg values being lost for nested includes

meld-cp:
Named argument values weren't being restored correctly if the same arg. name was being used for nested includes.

This PR includes a test case and a fix for the issue.


xoofx:
Yeah, it's a feature that I have never used so I'm not surprised that it has several issues. Thank you for investigating them.

PR #648, opened 2026-03-09
Fix `Include` resolve named argument values when StrictVariables is enabled

Benkei:
This PR fixes a bug that occurs when StrictVariables is enabled during the Include process.

I hope this fixes the problem correctly.


xoofx:
Could you elaborate what is it solving? What I see is that with strict variables, we silently discard a runtime error, which feels... counter-intuitive? What is the use case and problem that this PR is trying to solve?

Benkei:
Thanks for the review. I'll check it again.

Issue #661, opened 2026-03-23
String function in 7.0.0 have breaking nullable annotations


Mertsch:
```
public static string Capitalize(string text)
public static bool Empty(string text)
public static bool Whitespace(string text)
public static bool EqualsIgnoreCase(string text, string value)
public static string Md5(string text)
public static string Sha1(string text)
...
public static string PadLeft(string text, int width)
public static string PadRight(string text, int width)
public static string Base64Encode(string text)
public static string Base64Decode(string text)
```
Do support `null` as input. See
https://github.com/scriban/scriban/blob/a0f28c4616af87898e5e67e9c0c90dc9f8dc0fe1/src/Scriban/Functions/StringFunctions.cs#L256-L259

Yet I am still getting
```
Expression triggered 'Fail (<input>(241,30) : error : Argument `text` cannot be null for function `Empty`.)'.
```

So I am guessing that nullable annotation do have an effect on how functions are loaded. And if so, "we" might verify nullable on all functions?! I have only found this with `string.empty`, but my tests are insufficient.

PR #677, opened 2026-06-22
Adding nullable support to is_number function

Mertsch:
This used to work before nullable annotation were introduced, similar to https://github.com/scriban/scriban/issues/661

Note: I have updated the `420-math.txt` and `420-math.out.txt` but I could not get the tests to fail, even with clearly wrong `420-math.out.txt`. Are tests working as expected?!

### DIFF
--- a/src/Scriban.Tests/TestFiles/400-builtins/420-math.out.txt
+++ b/src/Scriban.Tests/TestFiles/400-builtins/420-math.out.txt
@@ -8,4 +8,5 @@
 true
 true
 false
+false
 4.24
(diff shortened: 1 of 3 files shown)


xoofx:
Fixed via 8872919995b8baae3008c2d3be6f0fad04e812ac

xoofx:
> Note: I have updated the 420-math.txt and 420-math.out.txt but I could not get the tests to fail, even with clearly wrong 420-math.out.txt. Are tests working as expected?!

The implementation is supporting `null`, it was just the method contract that didn't reflect its support.

Mertsch:
@xoofx Thanks. What I mean is, that I could put any random text into `420-math.out.txt` and the tests would still pass when I ran them from inside of visual studio.

xoofx:
> @xoofx Thanks. What I mean is, that I could put any random text into `420-math.out.txt` and the tests would still pass when I ran them from inside of visual studio.

Indeed, there was some inconclusive tests that were discarding the other tests. I fixed this and they are all running again. That's weird, I will have to double check this.

Thanks!

PR #667, opened 2026-04-07
Added DateTime.UtcNow Support

boslandj:
* Added a built in function for getting the current date time in utc.
* Updated docs and tests for built in date time functions.

PR #684, opened 2026-09-19
Fix TestUtcDateNow comparing utc_now against local time

tahakocal:
`TestUtcDateNow` renders `date.utc_now` but compares the result against `DateTime.Now`, so it fails whenever the local date differs from the UTC date. The assertions look like they were copied from `TestDateNow` when `date.utc_now` was added in #667.

Running the test here (UTC+3, shortly after local midnight):

```
Expected: "20 Sep 2026"
But was:  "19 Sep 2026"
   at Scriban.Tests.TestParser.TestUtcDateNow() in src/Scriban.Tests/TestParser.cs:line 397
```

CI never sees it because GitHub runners are UTC, but it fails for contributors in other timezones for part of every day.

`date.utc_now` itself is correct: `DateTimeFunctions.UtcNow()` returns `DateTime.UtcNow`, and nothing in the rendering path converts time zones. So only the test changes: it now compares against `DateTime.UtcNow`, and the local is renamed to `utcNow` to make the difference from `TestDateNow` obvious.

`TestDateNow` is left alone, since `date.now` is local and comparing it with `DateTime.Now` is right.

Tested with `dotnet test -c Release` in `src`:

| TZ | Before | After |
| --- | --- | --- |
| UTC | passes | passes |
| Asia/Istanbul (+3) | **fails** | passes |
| Pacific/Kiritimati (+14) | **fails** | passes |
| Etc/GMT+12 | passes | passes |

The full suite passes (1096 tests).


xoofx:
Reviewed locally: date.utc_now calls DateTime.UtcNow directly, whereas the old assertions used local time. A deterministic UTC+3 boundary check (2026-09-19 22:00 UTC versus 2026-09-20 local) confirms the mismatch without changing the machine timezone. The PR branch passes the full NUnit suite: 1096 passed, 84 skipped. Runtime behavior is unchanged; no documentation changes needed. The host timezone was not modified, so the submitted cross-timezone execution matrix was not independently repeated.

Issue #404, opened 2022-01-24
Give input format to date.parse


yogeshwar-b:
How to give input Date Format to date.parse method ? 

I have looked at the documentation but did not find it . 

Model : 
`{"dob":"20/01/2022 08:32:48 +00:00" }`

Scriban Template : 
```
{{~
	func formatDate
		ret (date.parse $0)| date.to_string '%Y-%m-%d %H:%M:%S'
	end
~}}


DOB {{formatDate model.dob}}
```

Actual Output :

`DOB 0001-01-01 00:00:00`

Expected Output : 

`DOB 2022-01-20 08:32:48`

How to get expected output ?

Sorry if this is wrong place to ask this question.
Thanks in Advance.

xoofx:
The function used to parse is define here:

https://github.com/scriban/scriban/blob/78de3af89c0c1bcca04bbb228df4008a077d1164/src/Scriban/Functions/DateTimeFunctions.cs#L250-L262

I haven't checked, but it could be that this doesn't handle well when the time passed as time offsets, so it might require to try to parse it multiple steps (e.g with `DateTimeOffset.TryParse` if `DateTime.TryParse` fails)

PR Welcome.

yogeshwar-b:
I see that the date parse is working fine if the context culture is 'en-GB' in line 257 in above mentioned code.

I dont have access to the Scriban render code. Only thing i can modify is Template file , can i set culture from the template file ?

xoofx:
> I dont have access to the Scriban render code. Only thing i can modify is Template file , can i set culture from the template file ?

You could see from the code above that it was not possible to pass a culture.

Anyway, I fixed it in 072aa23, it should be available in the upcoming`5.2.0+`

I won't do that always (to bring something that you could have tried to bring to this repo yourself), so yep, either if you see me one day, buy me a beer, or please, consider making a one-time donation through GitHub. 😉

PR #686, opened 2026-09-20
Fix date.parse not matching full month names with %B

tahakocal:
`%B` is documented as "Month in full form" and `date.to_string` formats it with `MMMM`, but the entry in the `Formats` table carries `"MMM"` as the pattern used for parsing, so `date.parse` with `%B` never matches a full month name:

```
{{ date.parse '05 January 2016' '%d %B %Y' }}
```

renders `01 Jan 0001`. `date.parse` returns `default(DateTime)` when parsing fails, so nothing is reported: the date is silently replaced by the zero date. A value written with `date.to_string '%d %B %Y'` can't be read back with the same pattern.

The parse column was added in #404 and the `%B` row looks like it inherited `"MMM"` from the `%b` row above it.

The fix is that one pattern: `"MMM"` → `"MMMM"`.

# Behaviour changes

Both follow from the corrected pattern, and both align with what the docs say `%B` means:

- `%B` no longer parses abbreviated month names. `{{ date.parse '05 Jan 2016' '%d %B %Y' }}` used to work by accident and now returns the zero date, silently, like every other failed parse. Such templates need `%b`, which is unchanged.
- `date.parse_to_string` also formats its output pattern through the same table, so `{{ '2016/01/05' | date.parse_to_string '%d %B %Y' }}` changes from `05 Jan 2016` to `05 January 2016`, matching `date.to_string '%d %B %Y'`.

If you'd rather keep `%B` accepting both forms, I'm happy to change it to try the full name first and fall back to the abbreviation.

# Test Plan

Added two cases to `440-date.txt`: parsing a full month name, and the `parse_to_string` output above. The parse case renders ` 01 Jan 0001` without the fix.

Compared the old and new build over a matrix of templates. Fixed:

| Template | Before | After |
| --- | --- | --- |
| `date.parse '05 January 2016' '%d %B %Y'` | `01 Jan 0001` | `05 Jan 2016` |
| `date.parse '05 JANUARY 2016' '%d %B %Y'` | `01 Jan 0001` | `05 Jan 2016` |
| `date.parse '05 janvier 2016' '%d %B %Y' culture:'fr-FR'` | `01 Jan 0001` | `05 Jan 2016` |
| `date.parse '05 Januar 2016' '%d %B %Y' culture:'de-DE'` | `01 Jan 0001` | `05 Jan 2016` |
| `to_string '%d %B %Y'` then `parse` with the same pattern | `01 Jan 0001` | `05 Jan 2016` |
| `'05 January 2016' \| date.parse_to_string '%Y' input_pattern:'%d %B %Y'` | `0001` | `2016` |

Unchanged: `%b` with abbreviated names in every culture tested, `%B` with a mismatched culture (still rejected), garbage, empty and null input, and `date.to_string '%B'` itself.

`TZ=UTC dotnet test src/Scriban.Tests/Scriban.Tests.csproj -c Release` passes (1096 tests). Checked on net8.0 and net10.0.

(`TestUtcDateNow` fails on this branch in any non-UTC timezone for an unrelated reason, fixed in #684.)


tahakocal:
While checking this I went through the rest of the `Formats` table. `%B` is the only entry whose parse pattern is a different token from its formatter, but a few others don't round trip for a different reason, so I left them out of this PR:

- `%e`, `%k`, `%l` format with a space pad (`" 5"`) and parse without one, so `to_string` output doesn't re-parse for single-digit values.
- `%x` and `%X` use the standard specifiers `d` and `T`. Those only mean "short date" / "long time" when they are the whole pattern, so embedded in a longer pattern such as `'%x %H'` they are read as a custom day specifier and a literal `T`.

Happy to follow up on those separately if you'd like them fixed.

xoofx:
Verified locally: the new golden date fixture fails with the master implementation and passes with the MMMM mapping. Full suite passes after integrating current master: 1096 passed, 84 skipped. The change restores the documented full-month meaning of %B and makes parse_to_string output consistent with that meaning. The disclosed compatibility change is intentional: abbreviated input should use %b. Existing format documentation already describes %B as full month; no documentation correction is needed. Ready to squash-merge.

PR #687, opened 2026-09-20
Fix string.truncatewords ignoring its ellipsis argument

tahakocal:
`string.truncatewords` takes an `ellipsis` argument, documents it in the remarks and in `site/docs/builtins/string.md`, and then never reads it: the method appends a hardcoded `"..."`.

```
{{ "This is a test truncated at 5" | string.truncatewords 5 " (...)" }}
```

renders `This is a test truncated...`.

`string.truncate`, right above it, handles the same argument correctly, so this looks like an omission rather than a decision.

The fix mirrors `Truncate`: default the argument with `ellipsis = ellipsis ?? "..."` and append it.

Liquid's `truncatewords` filter maps straight to this function (`LiquidBuiltinsFunctions.cs`), so `{{ "..." | truncatewords: 4, " --" }}` also starts honouring its second argument, which matches Shopify liquid.

Not changed here: the ellipsis is appended even when nothing was truncated (`{{ "a b" | string.truncatewords 4 }}` → `a b...`), unlike `string.truncate` and unlike liquid. That is a separate behaviour change, so I left it out of this PR. Passing `""` as the ellipsis works as an opt-out.

# Test Plan

Added golden cases for a custom ellipsis, for an empty ellipsis, and for the liquid filter with a custom ellipsis. The custom-ellipsis case renders `This is a test truncated...` without the fix.

`TZ=UTC dotnet test src/Scriban.Tests/Scriban.Tests.csproj -c Release` passes (1096 tests).


xoofx:
Verified locally: keeping the PR regression fixtures but restoring master implementation fails both the Scriban custom/empty ellipsis cases and Liquid fixture (2 failing tests). Restoring the fix passes the full suite, also after integrating current master: 1096 passed, 84 skipped. Existing XML/site documentation already promises the ellipsis argument; this restores that contract. The separately noted unconditional suffix behavior is unchanged and outside this fix. Ready to squash-merge.

PR #688, opened 2026-09-20
Fix date.parse_to_string formatting output with the parse pattern converter

tahakocal:
`date.parse_to_string` builds its output with `ParseCustomFormat`, the strftime to .NET-format-string converter used for *parsing*, and then calls `DateTime.ToString` with the result. The remarks say "See `to_string` method about the format of a pattern", but the output side doesn't behave like `to_string`:

- Literal text in the pattern is passed through to `DateTime.ToString` unescaped, so its letters are read as format specifiers: `{{ '2016/01/05' | date.parse_to_string 'Day %d of %B' }}` renders `Da16 05 o0 Jan`.
- With no output pattern, the raw default pattern `%d %b %Y` goes to `DateTime.ToString`, rendering `5 b Y`.
- Specifiers that have no parse mapping (`%j %s %C %u %w %U %V %W %% %n %t %v`) throw "The pattern %X is not supported for the parse method" when used as *output*, although `to_string` renders them.

# How

`ToString(DateTime?, string, CultureInfo)` already walks the pattern and calls the formatter for each specifier. That loop moves into a private `FormatDateTime` helper, and both `to_string` and `parse_to_string` use it, so the two can't drift apart again. `ParseCustomFormat` stays where it belongs, on the input pattern.

`to_string` keeps its own `%g`-only special case (expanding to `"%g " + Format`); `parse_to_string` falls back to the default pattern as before.

# Test Plan

Added golden cases for a pattern with literal text, for no output pattern, and for `%j`. Without the fix the `%j` case fails with the parse error and the other two render `Da16 05 o0 Jan` and `5 b Y`.

`TZ=UTC dotnet test src/Scriban.Tests/Scriban.Tests.csproj -c Release` passes (1096 tests), so the documented examples are unchanged.

Ran the old and new build over the same templates:

| Template | Before | After |
| --- | --- | --- |
| `parse_to_string '%Y--%m--%d'` (documented) | `2016--01--05` | `2016--01--05` |
| `parse_to_string "%b %d, %y"` (documented) | `Mar 14, 16` | `Mar 14, 16` |
| `parse_to_string "%b %d, %y" input_pattern: "%m %d, %Y"` (documented) | `Mar 14, 16` | `Mar 14, 16` |
| `parse_to_string 'Day %d of %B'` | `Da16 05 o0 Jan` | `Day 05 of January` |
| `parse_to_string` (no pattern) | `5 b Y` | `05 Jan 2016` |
| `parse_to_string '%%'` | error | `%` |
| `parse_to_string '%g %d %b %Y'` | ` 05 Jan 2016` | ` 05 Jan 2016` |
| `parse_to_string '%d %B %Y' output_culture:'fr-FR'` | `05 janv. 2016` | `05 janvier 2016` |
| `parse_to_string '%d %B %Y' output_culture:'en-US'` | `05 Jan 2016` | `05 January 2016` |
| null input, empty input | empty | empty |
| unparseable input | `0001` | `0001` |
| `date.to_string 'Day %d of %B'` | `Day 05 of January` | `Day 05 of January` |
| `date.to_string '%g%X'` | `21:22:23` | `21:22:23` |

The two `%B` rows change because the output pattern now uses the same formatter as `to_string`, which renders `%B` as the full month name. `date.to_string` itself is unaffected.


xoofx:
Verified locally: the new fixture fails against the master implementation and passes with the shared formatter. Reviewed the extraction: date.to_string retains its %g-only special case, while parse_to_string formats literals, default output, and output-only specifiers through the formatting path rather than the parse converter. Integrated master and resolved overlapping golden-fixture additions by retaining both sets of regression cases, then pushed the integration commit. Full suite passes: 1096 passed, 84 skipped. Existing docs already reference to_string output syntax. Ready to squash-merge.

PR #689, opened 2026-09-20
Fix array.sort failing on numbers of different types

tahakocal:
`array.sort` throws on a list that mixes CLR numeric types:

```csharp
var items = new ScriptArray { 3, 1.5, 2L, 0.5m };
ArrayFunctions.Sort(new TemplateContext(), new SourceSpan(), items);
// InvalidOperationException: Failed to compare two elements in the array.
// -> ArgumentException: Object must be of type Int32.
```

This happens with plain template data too, for instance an array built from a C# model whose members are `int` and `long`, or `[3, 1.5] | array.sort`. `Comparer<object>.Default` requires both values to be the same type, and `OrderBy` surfaces that as a failed sort.

# How

The comparer defines one ordering and uses it for every pair of numbers: NaN first, then negative infinity, then finite values, then positive infinity. Finite values are compared as exact fractions - integers as they are, `decimal` from `decimal.GetBits` (mantissa over `10^scale`), `float` and `double` from their IEEE-754 fields (mantissa over a power of two, subnormals included) - and the comparison is a single `BigInteger` cross-multiplication, so no value is rounded and no type is promoted.

Pairs of the same CLR type keep the existing `Comparer<object>.Default` path, which agrees with the exact ordering for every type `MathFunctions.IsNumber` accepts. Non-numeric values are untouched.

The first revision of this PR reused the language comparison (`ScriptBinaryExpression`). As @xoofx showed in review, that is not a valid ordering: `Compare(NaN, 1)` disagreed with `Compare(NaN, 1.0)`, and `9007199254740993L` compared equal to `9007199254740992d` while comparing greater than `9007199254740992L`. Both are fixed by comparing exact values instead.

Two visible consequences:

- Mixed `decimal`/`double` comparisons that used to throw when promotion overflowed (`decimal.MaxValue` against `1e40`) now compare exactly.
- `array.sort` is more precise than the `<` operator: `9007199254740993 < 9007199254740992.0` is still false in a template, while sort orders those two correctly.

# Test Plan

Regression tests for the original bug (`array.sort` over `int`/`double`/`long`/`decimal`, directly and through a member), and for the ordering itself: NaN placement, integer-versus-double precision, `0.1m` against `0.1`, `-0.0` against `0`, infinities across types, and a comparer-law test (reflexivity, antisymmetry, transitivity over 40 values covering every numeric type, subnormals, decimal scales, `BigInteger` beyond double range, NaNs and infinities).

Beyond the suite, every pair from that value set was compared against an exact-fraction oracle computed independently: 3,268,665 pairs, 0 disagreements; 82,225,043 triples, 0 transitivity or antisymmetry failures; 1,600 sorts of shuffled mixed-type lists, 0 exceptions. A naive "convert both to double" comparer gets 814 of those pairs wrong, so the set discriminates.

Full suite: 1111 passed, 84 skipped.

The ordering is documented in the `array.sort` remarks and `site/docs/builtins/array.md` was regenerated.


xoofx (changes requested):
Parking this PR for now: the underlying bug is real, but the proposed comparer is not a valid ordering for all supported numbers.

I verified both submitted regression tests fail against master and pass with this implementation. The submitted branch's full suite passes (1098 passed, 84 skipped), but two additional local checks fail:

1. **NaN ordering depends on the other operand's CLR type.** The new comparer returns `0` for `(double.NaN, 1)` and `0` for `(1, 1.0)`, but `-1` for `(double.NaN, 1.0)`. Equality in the comparer is therefore not transitive. The public sort of `new ScriptArray { 1, double.NaN, 0.5 }` produces `0.5, 1, NaN`, rather than the consistent NaN-first ordering of the default floating-point comparer. The description's claim that `Comparer<double>.Default` leaves NaN wherever it started is inaccurate; it orders NaN before ordinary numbers.

2. **Precision loss also violates transitivity without NaN.** For `a = 9007199254740993L`, `b = 9007199254740992d`, and `c = 9007199254740992L`, this comparer returns `Compare(a,b) = 0`, `Compare(b,c) = 0`, and `Compare(a,c) = 1`. Mixing language numeric promotion with exact same-type comparisons cannot provide the consistent ordering required by OrderBy.

These were executed locally against head `558f3ee63e4c016a32a22eaba7c9307e097159cc`, using the actual nested SortComparer via reflection and ArrayFunctions.Sort for the public sorting example. Minimal comparer probe:

```csharp
var type = typeof(ArrayFunctions).GetNestedType("SortComparer", BindingFlags.NonPublic)!;
var comparer = (IComparer<object?>)Activator.CreateInstance(
    type, new TemplateContext(), new SourceSpan())!;
// These assertions fail on this PR:
Assert.That(comparer.Compare(double.NaN, 1), Is.LessThan(0));
Assert.That(comparer.Compare(9007199254740993L, 9007199254740992d), Is.GreaterThan(0));
```

Before landing, this needs a deliberate, consistent mixed-numeric ordering (including NaN, infinities, large integers, decimals and BigInteger), with comparer-law/precision regression coverage through both direct and member sorting. A NaN-only patch would not address the second problem, and broadening the language comparison semantics is beyond a safe review follow-up. Keeping the `bug` label because the original issue is valid; leaving this open and unmerged with changes requested.

tahakocal:
You were right on both counts, and the description's claim about `Comparer<double>.Default` was wrong. Rewritten in 43fdcf2: the comparer no longer borrows the language comparison, it defines one ordering and uses it for every numeric pair.

**The ordering.** NaN, then negative infinity, then finite values, then positive infinity. Finite values are compared as exact fractions: integers as they are, `decimal` from `decimal.GetBits` (mantissa over `10^scale`), `float`/`double` from their IEEE-754 fields (mantissa over a power of two, subnormals included). The comparison is one cross-multiplication of `BigInteger`s, so nothing is rounded and no type is promoted. Pairs of the same CLR type keep the old `Comparer<object>.Default` fast path, which agrees with the exact path for all twelve types `MathFunctions.IsNumber` accepts (`Double.CompareTo` puts NaN below negative infinity and treats `-0.0` as equal to `0.0`, which is what the ranks and the exact values do).

Your two probe assertions now hold:

```
Compare(double.NaN, 1)                          -> -1   (was 0)
Compare(9007199254740993L, 9007199254740992d)   ->  1   (was 0)
```

And the example from your review sorts as you described:

```
[1, NaN, 0.5] -> NaN, 0.5, 1
```

**Transitivity.** The second problem is gone for the same reason: `Compare(9007199254740993L, 9007199254740992d)` is now `1`, so the `a > b = c` chain is consistent. I checked this rather than assuming it. Over a set built from every numeric type — `±0`, scale variants (`1.0m`, `1.00m`), `decimal.MaxValue`, `2^53±1`, `2^63`, `2^64±1`, `long.MinValue`, `ulong.MaxValue`, subnormals (`double.Epsilon`, `1e-310`, `float.Epsilon`), values that are not representable in decimal (`0.1`, `1e40`, `1e300`), `BigInteger` up to `10^400`, NaNs with different payloads and signs, and both infinities — I compared every pair against an exact-fraction oracle computed independently (Python `Fraction`, which is exact for `int`, `Decimal` and `float`):

- 3,268,665 pairs checked, 0 disagreements with the oracle
- 82,225,043 triples checked, 0 transitivity failures, 0 antisymmetry failures
- 1,600 sorts of shuffled mixed-type lists, 0 exceptions
- as a control, a naive "convert both to double" comparer gets 814 of those pairs wrong, so the set does discriminate

**Tests.** `TestSortComparerOrdersNaNFirst`, `TestSortComparerKeepsIntegerPrecisionAgainstDouble` (including `Compare(0.1m, 0.1) < 0` and `BigInteger.Parse("10000000000000000303786028427003666890752") == 1e40`, the exact value of that double), `TestSortComparerIsATotalOrder` (reflexivity, antisymmetry and transitivity over 40 values across all the families above), plus `TestSortOrdersNaNBeforeNumbers`, `TestSortOrdersInfinitiesAcrossNumberTypes` and `TestSortKeepsIntegerPrecisionAgainstDouble` through the public `array.sort`. Full suite: 1111 passed, 84 skipped.

Two consequences worth calling out, both visible changes:

- Mixed `decimal`/`double` comparisons that used to throw (promotion overflows, e.g. `decimal.MaxValue` against `1e40`) now compare exactly.
- `array.sort` is now more precise than the `<` operator: `9007199254740993 < 9007199254740992.0` is still false in a template, while sort orders those two correctly. That follows from the ordering you asked for; say the word if you would rather have sort match the operator instead.

The comparer is a stateless singleton now (it no longer needs `TemplateContext`/`SourceSpan`), so the probe from your review would reach it as `ArrayFunctions.SortComparer.Instance`. I also documented the ordering in the `array.sort` remarks and regenerated `site/docs/builtins/array.md`.

xoofx (approved):
Re-reviewed revision 43fdcf2. Both previous blockers are resolved: NaN ordering is consistent across numeric types, and exact mixed-numeric comparisons preserve large-integer precision and comparer transitivity. The exact decimal/IEEE-754 decomposition agrees with the same-type comparison path; nonnumeric/null behavior and stable sorting are preserved. Documentation matches the new ordering.

Local verification: Release solution build succeeded with zero warnings/errors; full suite passed (1111 passed, 84 skipped); all 12 sort tests passed. Additional review-only probes passed for 100 shuffled direct/member sorts spanning all 12 supported numeric types, null, NaN, infinities, subnormals, decimal boundaries, 2^53 precision boundaries, and BigInteger values beyond double range, plus stable member ordering across equivalent values of all 12 types. GitHub CI is also green. No required changes remain.

PR #690, opened 2026-09-20
Fix date.parse with %x and %X combined with other specifiers

tahakocal:
`%x` and `%X` parse correctly on their own but not when combined with anything else:

```
{{ (date.parse '01/05/2016' '%x').year }}                  -> 2016
{{ (date.parse '21:22:23' '%X').hour }}                    -> 21
{{ (date.parse '01/05/2016 21:22:23' '%x %X').year }}      -> 1
```

So a value written with `date.to_string '%x %X'` can't be read back with the same pattern, and the failure is silent, since `date.parse` returns the zero date when parsing fails.

`ParseCustomFormat` builds one custom format string out of the parse patterns in the `Formats` table. `%x` and `%X` map to `"d"` and `"T"`, which are *standard* format specifiers: they mean "short date" and "long time" only when they are the entire format string. Concatenated with anything else they are read as the custom specifiers for a single-digit day and a literal `T`.

# How

Expand those two to the culture's equivalent custom pattern (`ShortDatePattern`, `LongTimePattern`) when building the parse format, so they keep their meaning inside a longer pattern. Formatting is unchanged; it passes each specifier to `DateTime.ToString` on its own, where the standard form is correct.

# Test Plan

Added two golden cases: parsing `'%x %X'` and a round trip through `date.to_string '%x %X'`. Without the change they render `1` and ` 01 Jan 0001`.

`TZ=UTC dotnet test src/Scriban.Tests/Scriban.Tests.csproj -c Release` passes (1096 tests).

| Template | Before | After |
| --- | --- | --- |
| `(date.parse '01/05/2016 21:22:23' '%x %X').year` | `1` | `2016` |
| `date.parse (… \| date.to_string '%x %X') '%x %X'` | `01 Jan 0001` | `05 Jan 2016` |
| `(date.parse '01/05/2016' '%x').year` | `2016` | `2016` |
| `(date.parse '21:22:23' '%X').hour` | `21` | `21` |
| `date.parse '2016/01/05 21:22:23' \| date.to_string '%x %X'` | `01/05/2016 21:22:23` | `01/05/2016 21:22:23` |

Related, not fixed here: `%e`, `%k` and `%l` format with a space pad (`" 5"`) but parse without one, so their output doesn't round trip either. That needs a different fix, so I left it out.


xoofx:
Verified the original claim locally: the PR golden fixture fails with master implementation and passes with the expansion. Additional review found a culture regression: with fr-FR and pattern %x %X%g, 01/05/2016 was parsed as May 1 instead of January 5, because expansion used French patterns before the final invariant override. Pushed a fix that resolves the single parse culture before expanding patterns, plus 7 regression cases covering French/German patterns and %g before/between/after standard specifiers, including standalone %x. The two combined trailing/middle override cases failed before this follow-up and now pass. Updated XML/site docs, integrated master, and preserved all overlapping date fixtures. Release solution build: 0 warnings/errors. Full solution tests: 1103 passed, 84 skipped. Ready to squash-merge.

